PR #988, opened 2015-01-03
regenerate locales after updating mathjax-i18n

pkra:


dpvc:
I'm not sure I understand what this pull request does.  The changes all seem to be in the version numbers (removing the `-beta`, which shouldn't happen until we do the actual v2.5 release).

pkra:
Actually, there are a few actual updates hiding in the invisible masses... I'll close and create a clean one.

PR #1001, opened 2015-01-09
texvc to tex conversion table

physikerwelt:
@pkra are you still interested in texvc... I reimplemented everything from scratch really minimal
Please also check the licensing issue and the minified javascript… I could not find documentation on this is done for mathjax


pkra:
Yes definitely still interested in this (especially if it helps your work on MediaWiki/Wikipedia!).

Can you make the pull request to the `develop` branch, please? `master` is our stable release branch.

Would be ok to rename this to something like "MediaWiki.js"? I think people are more likely to understand what's behind that and it would make it easier for contributors to identify.

(Just FYI, we don't need packed versions; we repack for testing and releases anyway. And thanks for the demo file; our [test suite](http://github.com/mathjax/mathjax-test) is separate but for now we can proceed without tests).

unpacked/extensions/TeX/texvc.js
```diff
@@ -0,0 +1,95 @@
...
+			spades: ["Macro", "\\spadesuit"],
+			sub: ["Macro", "\\subset"],
+			sube: ["Macro", "\\subseteq"],
+			supe: ["Macro", "\\supseteq"],
+			Tau: ["Macro", "\\mathrm{T}"],
+			thetasym: ["Macro", "\\vartheta"],
+			uarr: ["Macro", "\\uparrow"],
+			uArr: ["Macro", "\\Uparrow"],
+			Uarr: ["Macro", "\\Uparrow"],
+			varcoppa: ["Macro", "\\mbox{\\coppa}"],
```
pkra: This needs fixing -- MathJax only parses some macros within text mode so `$\varcoppa$` would give something like `\coppa` as output.

Replacing `\\coppa`` with `\\u03D9` should be sufficient.

pkra: What about the rest from https://phabricator.wikimedia.org/T75173 ?

physikerwelt:
Changing to develop is not possible (see https://github.com/isaacs/github/issues/18)
So I close this pull request. See #1003 for the new pull request.

Issue #795, opened 2014-04-21
Angstrom symbol (\AA) does not render in MathJax


eteq:
It appears the [Angstrom symbol](http://en.wikipedia.org/wiki/Angstrom) is not rendering in MathJax, despite being quite common in LaTeX packages, and has a unicode character (U+212B , normalized to U+00C5 ).  Can this be added?  It's used in a variety of physical sciences.

See astropy/astropy#2286 ipython/ipython#5533 for the original trigger/use-case for this.

Issue #537, opened 2013-08-07
Tex Teubner set


hartman:
It might be nice to expand the current selection of Tex packages with parts of the teubner set of greek symbols.

Wikipedia's uses the following Teubner commands in their texvc, for use in articles containing ancient greek numerals and arithmetic:
\Coppa, \coppa, \varcoppa, \Digamma, \Koppa, \koppa, \Sampi, \sampi, \Stigma, \stigma, \varstigma
Some are additions, some are redefines of AMS.

http://archive.cs.uu.nl/mirror/CTAN/macros/latex/contrib/teubner/teubner-doc.pdf
There might be some fontsupport work in this, due to changed glyphs over the course of history....

Related to: https://gerrit.wikimedia.org/r/#/c/4060 and https://bugzilla.wikimedia.org/35186

PR #1003, opened 2015-01-09
MediaWiki-texvc support

physikerwelt:
We add a conversion table that converts MediaWiki specific
markup to standard LaTeX. This approach is more or less
maitenance free, and relies on the existence some special
macros mostly originating from amsmath and amsfonts package.

In the future one could expand the macros with semantic annotation
like for instance that $\\mathbb{R}$ stands for the semantic
concept reals, if it was produced using the texvc macro $\\reals$.


unpacked/extensions/TeX/mediawiki-texvc.js
```diff
@@ -0,0 +1,95 @@
...
+			spades: ["Macro", "\\spadesuit"],
+			sub: ["Macro", "\\subset"],
+			sube: ["Macro", "\\subseteq"],
+			supe: ["Macro", "\\supseteq"],
+			Tau: ["Macro", "\\mathrm{T}"],
+			thetasym: ["Macro", "\\vartheta"],
+			uarr: ["Macro", "\\uparrow"],
+			uArr: ["Macro", "\\Uparrow"],
+			Uarr: ["Macro", "\\Uparrow"],
+			varcoppa: ["Macro", "\\mbox{\\coppa}"],
```
pkra: [From the previous PR.] This needs fixing -- there's no `\coppa` macro. Also, MathJax only parses some macros within text mode so $\varcoppa$ would give something like `\coppa` as output. Replacing `\\coppa` with `\\u03D9` should be sufficient?

(BTW, what about the rest from https://phabricator.wikimedia.org/T75173? I remember in the past there was `\AA` as well?)

physikerwelt: Can't we fix that somewhere else, or ignore this for now.
Currently this file is the exact definition for MediaWiki specific macros.
Accoding to my statistics \varcoppa is not used anywhere in enwiki.
Would you be happy to uncomment that until #537 is resolved, or just live with the fact if sombody will eventaully use it a see a \\coppa.

\AA seems to be standard tex and not Mediawiki specific. See for example
https://molspect.chemistry.ohio-state.edu/symposium/latexinstruct.html

pkra: > Can't we fix that somewhere else, or ignore this for now.
> Would you be happy to uncomment that until #537 is resolved, or just live with the fact if sombody will eventaully use it a see a \coppa.

Uncommenting would be fine. I guess we could also change it while merging it. This is a bug on the one end or the other -- so maybe consider it a first bug fix in this new setup :-) When we accept `varcoppa`, we're basically saying we acept a temporary/hacky solution for #537 (so there's an argument to do it for all of #537). I would be ok with that b/c it's MediaWiki.

> \AA seems to be standard tex and not Mediawiki specific

Technically, it's a standard *text mode* macro and in math mode, you'll need an additional TeX package like sinuitx (though technically you'll get a warning and a bad rendering these days). See the discussion at #795. Again, I'd be open in this context but we should make sure that this extension contains everything that MathJax doesn't support in any other extension.

physikerwelt: @pkra I think you have the best overview of the MathJax limitations. Depending on the settings of the MediaWiki installation, MathJax will receive either $\varcoppa$ or the texvjc preprocessed expression $\\mbox{\\coppa}$, therefore I had the feeling that it would be easiest to just expand the MediaWiki specific macros with their defintions. I think this is a well described initial commit, for this purpose. I would see improvements to that as follow ups.
@pkra Please feel free to change/add whatever you think needs to be changed.

pkra:
Ok, I took a closer look.

Already in MathJax are 

```
Doteq
doublecap
doublecup
ge
gggtr
le
ne
omicron
restriction
```

So we can drop those.

The following macros are currently available in the MW math extension but not in this extension.

@physikerwelt Could you confirm that they should be dropped here?

```
AA
coppa
Coppa
koppa
Koppa
oiiint
oiint
P
pagecolor
stigma
Stigma
textvisiblespace
vline

\\ set to undefined
arccot
arccsc
arcsec
sgn

\\ deactivated MathJax macros
style
href
```

pkra:
To sum up F2F discussion, I will pull this into my local develop (to retain the commit history) and finish it up before pushing it to develop. Hope to get to this tomorrow. (closing this for now)

Issue #963, opened 2014-11-26
v2.5 final bug


pkra:
* [x] finish release checklist
* [x] update documentation
* [x] create `2.5-latest` copy on CDN
* [x] update `latest` branch CDN.
* [x] merge final pull request from `develop` into `master`

pkra:
@dpvc I've created https://github.com/mathjax/MathJax/wiki/Draft-for-what%27s-new-page in case you want to look it over.

dpvc:
A couple of comments:

* I would not include issue 970 or 975 as they are not bugs in 2.4, but bugs that were found in the alpha of 2.5

* For 900, I'd say "Fix `\buildrel` to get TeX class REL instead of ORD" to make it clearer (it sounds like it might be the reverse).

* In 375, there is an error in escaping ``\}``.

* In 982, there is an escaping problem in the Markdown.

* In the AsciiMath section, do you want to mention the update to v2.2 of AsciiMathML?  I know there are other improvements not listed in our tracker (but in the AsciiMath one).

Other than that, looks good.

pkra:
Thanks, Davide. I've updated the doc.

pkra:
Completed.

PR #1027, opened 2015-01-29
Release v2.5

dpvc:
This is the pull request for the official v2.5 release.  It has been packed and combined, and the copyrights are updated.  I have smoke tested, and it looks good.  I will run a full test suite in each output mode for one brewer just to be sure, but it looks good at the moment.


dpvc:
The test suite runs went great.  I used the packed versions, so that gives not only testing of the final copy of v2.5, but also of the packed version (the main test runs use the unpacked version).  This should also mean that the tests ran with the CHTML-preview, so that provides a test of the two-pass process that isn't currently in the test suite otherwise.  So everything looks good to go.

pkra:
Fixes #963

PR #1040, opened 2015-02-03
update `mathjax-node` branch to MathJax v2.5 final

pkra:


pkra:
I did a smoke test and didn't spot any issues on mathjax-node. => ready for review.

dpvc:
There is an issue that this pull request brings up that we may want to consider.  It appears that you copied files from v2.5 into this branch and committed them rather than merging v2.5 into this branch and removing unwanted files.  This means that you lose the history for the updated and new files and the connection between this branch and v2.5 is not clear.  From the standpoint of functionality, there is no difference, but from a maintenance standpoint, it makes this branch somewhat more mysterious.

I am wondering if it wouldn't be netter to merge `v2.5-latest` or `master` into `MathJax-node` and then remove the unneeded files (I'm not certain they would actually be added back in).  That would make a clear indication of the connection to the current version of MathJax in the network diagram, and the files would maintain their update history and other meta-data.  You could probably use grunt to do the removals.

What do you think?

pkra:
I had tried that but ran into trouble at some point. I'll try again.

dpvc:
The deleted files do seem to be problematic.  Perhaps there is a different merge strategy that can be used to automatically resolve those in favor of the MathJax-node branch?

pkra:
How about I delete the branch here and create a freshly pruned version off v2.5-latest? 

I don't see a workable merge strategy after doing some more research. We could revisit the problem of splitting up the core repository, I suppose.

pkra:
With "delete the branch" I mean the mathjax-node branch.

dpvc:
I was thinking that might be the best approach as well.  The deleted files issue does seem to be a problem for git.

pkra:
Ok with me. In that case, we can't do any review though as I'd have to push directly to the "new" mathjax-node branch.

dpvc:
I already have looked at the mj25-node and it looks good, so as long as you reproduce that, it should be OK.  Otherwise, you can leave the current MathJax-node for now and recreate mj25-node as a branch from v2.5-latest, and we can review that.  Then delete the current MathJax-node and either rename mj25-node to mathjax-node, or create a new mathjax-node (from v2.5-develop) and merge mj25-node into it as a fast-forward merge.

pkra:
Ok, I've created a new branch `mathjax-node-2.5` from `v2.5-latest`. Pull request didn't seem to make a whole lot of sense to any existing branch. Could your review it?

dpvc:
Looks good.  I ran a bunch of files with various input and output formats.  The CommonHTML output jax and CHTML-preview are not really needed, as they are not currently used by MathJax-node, but they are not all that big, so don't matter too much.  CommonHTML eventually will be available.

dpvc:
I think it might be good to keep the name as `mathjax-node-2.5` (and change the `package.json` file in MathJax-node), as that provides some versioning in the MathJax-node's usage of MathJax.

pkra:
> The CommonHTML output jax and CHTML-preview are not really needed, as they are not currently used by MathJax-node

Do you mean that as in "not used in the examples in /bin"? CHTML worked fine when I tried it out a while back.

> I think it might be good to keep the name as mathjax-node-2.5

Ok.

dpvc:
> Do you mean that as in "not used in the examples in /bin"? CHTML worked fine when I tried it out a while back.

Sorry, forgot that you could specify any output renderer in pj-page.  I was thinking about mj-single, which doesn't have that as an option.  OK, leave them in.  (I still don't see any use for the CHTML-preview, but it is small, so no real need to remove.)

pkra:
> I still don't see any use for the CHTML-preview, 

Ah, right. I'll update the configuration in mathjax-grunt-clearner to include it for the hotfix.

PR #1047, opened 2015-02-05
remove README-branch.txt

### LINKED ISSUE #1039: README-branch.txt in master is outdated
Do we need it?

pkra:
Fixes #1039

### DIFF
--- a/README-branch.txt
+++ b/README-branch.txt
@@ -1,1 +1,0 @@
-This is release branch v2.4 of MathJax.


dpvc:
OK, this is closer, but the `issue1039` branch still branches from `develop`.  It should branch from `2.5-latest`.  I know there is not much difference at this point, but if we are to follow the process correctly, it is best if we do it properly right from the beginning.

Since I already have several v2.5 patches that need to go into a hot fix, should I just do this one as one of those, and not worry about a separate pull request?

pkra:
> It should branch from 2.5-latest

I didnt' see it as a hotfix issue so `develop` seemed natural to me.

> should I just do this one as one of those, and not worry about a separate pull request?

Sounds good.

dpvc:
Since it's wrong in the master branch (and so misleading), and we're making a hotfix anyway, seems like it should be done.

Issue #1050, opened 2015-02-09
NativeMML tweaks on WebKit


pkra:
For when we get a chance to revisit the NativeMML tweaks for WebKit. Some tweaks now have the adverse effect.

Some examples:

1) Stretchy parentheses

```html
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
        <mrow>
          <mo>(</mo>
          <mtable>
            <mtr>
              <mtd columnalign="center">
                <mrow>
                  <mo>(</mo>
                  <mtable>
                    <mtr>
                      <mtd columnalign="center">
                        <mi>a</mi>
                      </mtd>
                      <mtd columnalign="center">
                        <mi>b</mi>
                      </mtd>
                    </mtr>
                    <mtr>
                      <mtd columnalign="center">
                        <mi>c</mi>
                      </mtd>
                      <mtd columnalign="center">

                        <mi>d</mi>

                      </mtd>
                    </mtr>
                  </mtable>
                  <mo>)</mo>
                </mrow>
              </mtd>
              <mtd columnalign="center">
                <mrow>
                  <mo>(</mo>
                  <mtable>
                    <mtr>
                      <mtd columnalign="center">
                        <mi>e</mi>
                      </mtd>
                      <mtd columnalign="center">
                        <mi>f</mi>
                      </mtd>
                    </mtr>
                    <mtr>
                      <mtd columnalign="center">
                        <mi>g</mi>
                      </mtd>
                      <mtd columnalign="center">
                        <mi>h</mi>
                      </mtd>
                    </mtr>
                  </mtable>
                  <mo>)</mo>
                </mrow>
              </mtd>
            </mtr>
            <mtr>
              <mtd columnalign="center">
                <mn>0</mn>
              </mtd>
              <mtd columnalign="center">
                <mrow>
                  <mo>(</mo>
                  <mtable>
                    <mtr>
                      <mtd columnalign="center">
                        <mi>i</mi>
                      </mtd>
                      <mtd columnalign="center">
                        <mi>j</mi>
                      </mtd>
                    </mtr>
                    <mtr>
                      <mtd columnalign="center">
                        <mi>k</mi>
                      </mtd>
                      <mtd columnalign="center">
                        <mi>l</mi>
                      </mtd>
                    </mtr>
                  </mtable>
                  <mo>)</mo>
                </mrow>
              </mtd>
            </mtr>
          </mtable>
          <mo>)</mo>
        </mrow>
      </math>
```

[...]

Issue #1033, opened 2015-01-31
\genfrac fails with some delimiters (e.g., <>)


PeterLuschny:
Hi!

    \begin{equation}
    \sum_{k=0}^n (-1)^k \genfrac < > {0pt}{}{n}{k} = \frac{4^{n+1}-2^{n+1}}{n+1} B_{n+1},
    \end{equation}

worked until yesterday, today we have

    <math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
      <merror>
        <mtext>\begin{equation}
    \sum_{k=0}^n&#xA0;(-1)^k&#xA0;\genfrac&#xA0;&lt;&#xA0;&gt;&#xA0;{0pt}{}{n}     {k}&#xA0;=&#xA0;\frac{4^{n+1}-2^{n+1}}{n+1}&#xA0;B_{n+1},
    \end{equation}</mtext>
      </merror>
    </math>

Regards
Peter Luschny

CharlesAnifowose:
Are you using the CDN `//cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML` ?

PeterLuschny:
I was using "http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML".

I switched now to "2.4-latest". The full configuration can be seen on
http://luschny.de/math/zeta/The-Bernoulli-Manifesto.html

CharlesAnifowose:
We've pulled back to `2.2-latest` instead of `latest` and we'll be monitoring to see if the issue persists.

Unfortunately, when MathJax fails like this, it does not seem to provide any mechanism for throwing an error, so we have to rely on people using the site telling us. Not ideal.

dpvc:
@CharlesAnifowose, do you mean `2.4-latest` rather than `2.2-latest`?  I'm not sure the reason to go back that many versions.  Also, can you try `2.5-latest` and see if that helps?  If it is, indeed, a caching issue, that would resolve it while still giving you the improvements from 2.5.

@PeterLuschny, I have figured out the issue in your case, and it is a bug in v2.5.  There were changes to `\genfrac` to get the sizing of the delimiters right, and the conversion of the delimiters to their unicode values was being done too early for the new arrangement.  Here is a temporary fix for now until we push an update.  Add the following

``` html
<script type="text/x-mathjax-config">
if (MathJax.version === "2.5.0") {
[...]

Issue #1032, opened 2015-01-31
[Math Processing Error]


karna77:
Hello,

The below math has rendered perfectly in firefox and chrome yesterday, but today we are getting 'Math processing error'. We are unable to find the issue behind this. Could you please shed light on this?

```
[...]

PR #1065, opened 2015-02-19
Hotfix 2.5

dpvc:
This fixes a number of small problems with v2.5, including #1033, #1032, #1006, #1039, and #1050.  Also, it recovers some of the speed gain in HTML-CSS that was lost in a change during the testing of v2.5 to resolve an issue there.    Finally, it fixes some problems with maction.

@pkra, are you ready to do an update to the CDN for v2.5.1?  If so, I will repack and recombine, then we can merge this and tag a new release.

I ran several broswer/OS combinations through the test suite on this branch, to make sure there were no unexpected results, and it looks good.

Did you intend the update to mml3 to be part of this hotfix?  Currently, it is in the develop branch, not the hotfix branch, so it is not in this merge right now.  If you want it to be included, I will cherry pick that commit from develop.


pkra:
> Did you intend the update to mml3 to be part of this hotfix? Currently, it is in the develop branch, not the hotfix branch, so it is not in this merge right now. If you want it to be included, I will cherry pick that commit from develop.

Yes, that would be great.

dpvc:
I have included your changes to mml3, as well as the fix for issue #1053 that was ready but hadn't been merged.  I repacked and recombined, so this is ready to go.  When you are ready to update the CDN, this can be merged into `master` and tagged as a release.  We will also need to make the corresponding pull from `hotfix-2.5` into `v2.5-latest`;  I can make a pull request for that here, if you like, or can do it by hand once this pull is made.  Once those are done, I will merge into develop by hand.

There are only a few of the files (41) that are changed, so I'm hoping you can just move those files to the CDN, rather than updating and purging everything.

This update does include a change to the HTML-CSS output that requires a change to the noErrors extension, so if the CDN updates one before the other, it might cause temporary processing errors like before.  Doing the noErrors extension first, and then the HTML-CSS jax.js file would be the order that would make it work best.  Those sites using a combined configuration file will get compatible versions automatically (since they are both in the same file), but those who roll their own configurations could experience a problem if the CDN doesn't get both new versions at the same time.  Let me know if you need more details.

pkra:
I won't make it today. Let's do the release Monday or Tuesday.

dpvc:
Sounds good.  Nothing critical in the hotfix.

pkra:
Should the MathJax version be bumped to 2.5.1?

unpacked/MathJax.js
```diff
@@ -46,8 +46,8 @@ if (window.MathJax) {window.MathJax = {AuthorConfig: window.MathJax}}
 // MathJax.isPacked = true; // This line is uncommented by the packer.
 
 MathJax.version = "2.5.0";
```
pkra: To clarify, this is what I meant with "MathJax version" in my earlier message. Should we bump this to 2.5.1?

dpvc: Yes, I guess I missed that.  I'll take care of it.

dpvc:
There is one more thing we might want to add to this.  The change to remove measurements in HTML-CSS output (for speed) has had adverse affects when unknown characters are used (which used to be measured, but now aren't, so the widths are wrong).  This affects both unknown characters, and characters in other fonts.  See this [stack exchange post](http://stackoverflow.com/questions/28706543/mathjax-overarrow-too-short) for an example.  We could special case the unknown characters without too much trouble (I think).  Since this is a new problem with v2.5, it would fit in with a hot fix for 2.5.  I'm inclined to want to fix it if we can.

In the meantime, you could amend your answer on SE to indicate the setting `noReflows:false` in the HTML-CSS block of the configuration would fix the problem the OP was having.

pkra:
> Since this is a new problem with v2.5, it would fit in with a hot fix for 2.5. I'm inclined to want to fix it if we can.

Sounds good to me.

dpvc:
OK, I've fixed the code so that unknown characters are measured even when `noReflows` is true.  This should resolve the problem in the SE posting.

I've done some smoke testing, but didn't rerun the testsuite, as the change was minor.  I did check that it doesn't cause unnecessary reflows.

dpvc:
So I think the branch is ready to go.  Let me know when you do the update, and I'll pull it into `v2.5-latest` and `develop` as well.

pkra:
I've just updated GCS and purged CloudFlare. Changes are coming through on my end.

pkra:
@dpvc I created https://github.com/mathjax/MathJax/milestones/MathJax%20v2.5.1%20hotfix%20release -- did I miss anything that we should add there?

dpvc:
Here's some information from my end related to our talk yesterday about the cache breaking.  I am getting `MathJax.js` v2.5.1, but the copy of `TeX-AMS-MML_HTMLorMML-full.js` is still the v2.5.0 version (it contains none of the 2.5.1 files), even though the browser used  `?rev=2.5.1` to get it.  So It looks like the edge nodes are not asking for the new file due to the CGI parameters.  That only breaks the browser cache (which is what it was originally intended to do).  So purging of the changed files seems also to be necessary.

At this point, however, the browser cache will have the v2.5.0 file associated with the `?rev=2.5.1` since that's what the CDN served when the browser asked for it.  So I'd have to force the refresh to get the new version (when it becomes available).

Just information to keep in mind as we do such updates.

pkra:
Darn. Thanks for checking. I've initiated another purge which hopefully helps.

dpvc:
I'm still getting the 2.5.0 config files from both mathjax/latest and mathjax/2.5-latest.  But I do get the v2.5.1 files if I ask for the individually.  Did you purge the config directory?

dpvc:
The milestone looks correct.

dpvc:
I notices that Math.StackExchange is still using the beta copy.  We might want to suggest that they move back to the official release.

pkra:
OK. I had used an incorrect wildcard while updating GCS (but confusingly it partially did what I wanted; weird).

I've updated GCS and purged Clouflare. Third time's the charm?

pkra:
Forget what I just wrote -- the files are wrong in my local copy. I have no idea what happened.

pkra:
Oh nevermind. They're alright. I just got confused by the TeX input still being at 2.5.0.

pkra:
Are they coming through on your end?

dpvc:
If I use the address without the `rev=2.5.1` I get the correct version.  With the `rev=2.5.1` I get the old version.  This sounds somewhat familiar.  I'm wondering if the "aggressive" caching is the correct setting on the CDN?

dpvc:
I get the correct version for both on the `2.5-latest` branch, but not on the `latest` one.  This should not be this hard!

pkra:
I think I spotted a problem with gsutil -- it's not uploading everything... Investigating.

pkra:
Ok. I checked the files on GCS and it was only the configs that were somehow not uploaded (fun fact: gsutil throws `CommandException: Wrong number of arguments for "cp" command.` on those files. I have no idea why that would be...)

Are all changes coming through on your end now as well?

dpvc:
No, I'm still getting the strange behavior on `mathjax/latest` where I get the new version without `rev=2.5.1` but get the old version with it.  It seems that the `rev=2.5.1` version has been cached by the edge node I'm using from when the old config file wasn't yet replaced.  So now I'm getting that all the time.  That is why I'm wondering if "aggressive" is the right caching setting.  I don't know if there is a way for you to purge that.

I'm wondering if we should remove or change the revision number in MathJax.js so that we avoid those cached copies, now that the files have made it to the CDN properly.

pkra:
Is there a particular file where you get this?

dpvc:
It is [cdnVersion](https://github.com/mathjax/MathJax/blob/master/unpacked/MathJax.js#L50) in MathJax.js.  I would suggest making this something like `2.5.1x`.  It is used only for the `?rev=x.y.z` string, so changing it to something else will mean that none of the cached versions on the CDN will be used.

pkra:
As per our call, I'm refraining for now since the CDN seems to have normalized.

PR #1094, opened 2015-03-14
Create sample interactive example

alexan0308:
Hi! Exploring the possibility of your library MathJax. I like it very much. Had the idea to come up with an example of interactivity. I would be glad if you enjoy his example and you add it. Of course, at your sole discretion.

### DIFF
--- a/test/sample-interactive.html
+++ b/test/sample-interactive.html
@@ -1,0 +1,263 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>MathJax Test Page interactive equations</title>
+<!-- Copyright (c) 2009-2014 The MathJax Consortium -->
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+<meta http-equiv="X-UA-Compatible" content="IE=edge" />
+
+<script type="text/x-mathjax-config">
+  MathJax.Hub.Config({
+    extensions: ["tex2jax.js"],
+	messageStyle: "none",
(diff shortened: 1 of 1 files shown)


pkra:
That's a nice example! I'm not sure it fits in the core repo though -- @dpvc what do you think?

I'd say add it to https://github.com/mathjax/MathJax-examples but that repo hasn't attracted much attention. If you just want more people to see it, I'd be happy to mention it in the next community news on our blog (and you could put it wherever you like, e.g., gist, codepen) .

alexan0308:
ok, http://jsfiddle.net/alexan0308/0cuhbk1u/8/
Maybe you will be able to improve my code more professional or come up with a similar example. Please keep me informed.

>but that repo hasn't attracted much attention
why do you have on the site a few examples? may expose the site folder test? ))

dpvc:
It is a nice example, but I agree with Peter that it probably doesn't belong in the core test folder.  MathJax-exampes is probably better.

As for why we have examples in the core, originally (before we had the CDN) you had to install MathJax locally, and these were there to test your installation.  Then as the API grew more sophisticated, they were to illustrate some of the features and techniques.  This was before MathJax-examples existed.  So it is mostly for historical reasons, not a particular decision to have two sets of examples.

pkra:
Sorry, I forgot to close this when @alexan0308 opened https://github.com/mathjax/MathJax-examples/pull/1

PR #1100, opened 2015-03-17
Removing PNG from MathMenu

### LINKED ISSUE #1088: Menu icons improvements
Triggered by https://github.com/readium/SDKLauncher-iOS/issues/49

A few thoughts (that I apparently never posted on the tracker before)

* the icons do not show on IE8 (or IE6 for that matter)
* the icons do not show on Android Chrome (last I checked; the menu doesn't work well anymore)
* the icons force extra HTTP requests
* the icons make single-file builds slightly difficult 

So my propsal would be:

* switch to base64 encoding or straight to inline SVGs.

How does that sound?

pkra:
To fix #1088.

Very much work in progress. Just a quick first pass using Unicode points which have a good chance to work / be found everywhere (even old IE / WinXP and stock browser on Android 2.3).

* switch from `CloseX-31.png` to `u00D7` (i.e., `&times;`)
* switch submenu indicators to `u25BA`/`\u25C4` (dingbats)
* to do: clean up, move inline-styles to class, work on MathEvents.

@dpvc what do you think of this approach?


pkra:
@dpvc ready for review. (I couldn't figure out the MathEvents code so I just moved to base64 there)

dpvc:
I've made some suggested changes to your version, which you can see in [this comparison](https://github.com/mathjax/MathJax/compare/issue1088...dpvc:issue1088). 

The main impact of the change is to add backgrounds and borders to the "X" more in line with what the current icons look like.  While this doesn't matter too much for the About box and the Help dialog, it is important for the mobile menu close buttons, since they would otherwise appear floating off to the upper left, making them hard to see and difficult to use.  I also replaced the icon in MathEvents with a text-based one as well.  Finally, I added a color change on hover in order to make them "discoverable" (something that would have been more difficult with images).

If you like these changes, I can make the comparison above into a pull request, and it can be merged into this branch, and then we can merge that into develop.

pkra:
Sure.

PR #1141, opened 2015-03-27
 disable CHTMLpreview.js on IE<8

### LINKED ISSUE #1137: CHTMLpreview should detect IE<8 [was: MathJax 2.5 has broken IE 6/7 compatibility on Windows XP/Vista (v2.4 works, Win 7+ okay)]
Math is getting replaced with _[Math Processing Error]_'s.

IE 8 on XP/Vista/7+ is fine. Problem is occurring on http://math.stackexchange.com/questions as well as my own site.  I previously had a problem with IE 6-8 in v2.3 which got a speedy fix: mathjax/MathJax#672

Problems on: IE 6 XP; IE7 XP; IE7 Vista. IE8+ is unaffected.

pkra:
Fixes #1137

### DIFF
--- a/unpacked/extensions/CHTML-preview.js
+++ b/unpacked/extensions/CHTML-preview.js
@@ -26,7 +26,7 @@
  *  limitations under the License.
  */
 
-(function (HUB,HTML) {
+(function (HUB,HTML,BROWSER) {
   
   var SETTINGS = HUB.config.menuSettings;
 
@@ -42,7 +42,7 @@
       color: "inherit!important",
       updateTime: 30, updateDelay: 6,
       messageStyle: "none",
-      disabled: false
+      disabled: ((BROWSER.isMSIE && BROWSER.version < 8) ? true :  false)
     }),
 
     //
@@ -133,7 +133,7 @@
   
   HUB.Startup.signal.Post("CHTML-preview Ready");
 
-})(MathJax.Hub,MathJax.HTML);
+})(MathJax.Hub,MathJax.HTML,MathJax.Hub.Browser);
 
 MathJax.Ajax.loadComplete("[MathJax]/extensions/CHTML-preview.js");
 
--- a/unpacked/extensions/CHTML-preview.js
+++ b/unpacked/extensions/CHTML-preview.js
@@ -42,7 +42,7 @@
       color: "inherit!important",
       updateTime: 30, updateDelay: 6,
       messageStyle: "none",
-      disabled: ((BROWSER.isMSIE && BROWSER.version < 8) ? true :  false)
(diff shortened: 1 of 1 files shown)


dpvc:
I want to look at IE < 8 and see what the problem is. It might be something small that can be worked around rather than disabling entirely.

unpacked/extensions/CHTML-preview.js
```diff
@@ -42,7 +42,7 @@
       color: "inherit!important",
       updateTime: 30, updateDelay: 6,
       messageStyle: "none",
-      disabled: false
+      disabled: ((BROWSER.isMSIE && BROWSER.version < 8) ? true :  false)
```
dpvc: There is a function for checking version numbers, since versions can be x.y or x.y.z, and are strings, numeric checks for those are not correct (e.g., `BROWSER.version < 8.2` would be true if the version is 8.10 if you do a numeric check).  So I'd recommend `!BROWSER.versionAtLeast("8.0")` rather than `BROWSER.version < 8`, here.

thomastan:
Thanks! Look forward to seeing the fix.

dpvc:
See [my comment](https://github.com/mathjax/MathJax/issues/1137#issuecomment-88871217) on issue 1137.

Should the patch (either yours here or the one I suggest) be a hot-fix?  If so, then it should branch from master and be named `hotfix-2.5.2`, and the pull request would be to `master`.  You could still use an issue1137 branch if you want, but it should be branched from `hotfix-2.5.2` and the pull should be targeted to `hotfix-2.5.2` (with a separate pull from `hotfix-2.5.2` to master).  That would allow us to add more fixes, if needed.  (There are a couple of measurement issues in HTML-CSS that would be good to fix in a hot-fix as well.)

unpacked/extensions/CHTML-preview.js
```diff
@@ -42,7 +42,7 @@
       color: "inherit!important",
       updateTime: 30, updateDelay: 6,
       messageStyle: "none",
-      disabled: false
+      disabled: ((BROWSER.isMSIE && !BROWSER.versionAtLeast("8.0")) ? true :  false)
```
dpvc: At this point, the test can be simply:

    disabled: BROWSER.isMSIE && !BROWSER.versionAtLeast("8.0")

pkra: Doh. Thanks!

dpvc:
OK, now that issue 1137 is in the v2.5.2 milestone, which is a hotfix, this pull request will need to be made from the `hotfix-2.5.2` branch (now available).  I can rebase this one to that branch if you like (I've already been doing that with my branches that needed to be moved).

pkra:
Sounds good to me.

dpvc:
Closing this in favor of #1165 (hotfix pull request).

PR #1144, opened 2015-03-29
Now recheck when changes made while typesetting

### LINKED ISSUE #1132: Sample handles quick changes to large content badly
sample-dynamic-2.html ignores changes to the content if its currently rendering the mathjax. I largely copy-pasted the code into my own tool and found this to be a bit of an annoyance, resulting in me adding spaces to see changes, or typing and deleting quickly. Also doesn't really make sense since there's a queue being used.

jaredwindover:
Added a flag to signal that the content has changed while the typesetting was happening. When the typesetting is finished, another callback checks the flag and either finishes, or else starts typesetting again. This can close #1132.

### DIFF
--- a/test/sample-dynamic-2.html
+++ b/test/sample-dynamic-2.html
@@ -17,7 +17,7 @@
 <script>
 var Preview = {
   delay: 150,        // delay after keystroke before updating
-
+  
   preview: null,     // filled in by Init below
   buffer: null,      // filled in by Init below
 
@@ -58,7 +58,7 @@ var Preview = {
     if (this.timeout) {clearTimeout(this.timeout)}
     this.timeout = setTimeout(this.callback,this.delay);
   },
-
+  
   //
   //  Creates the preview and runs MathJax on it.
   //  If MathJax is already trying to render the code, return
@@ -83,7 +83,7 @@ var Preview = {
       );
     }
   },
-
+  
   //
   //  Indicate that MathJax is no longer running,
   //  and swap the buffers to show the results.
@@ -92,7 +92,9 @@ var Preview = {
     this.mjRunning = this.mjPending = false;
     this.SwapBuffers();
   }
+  
 };
+
 //
 //  Cache a callback to the CreatePreview action
 //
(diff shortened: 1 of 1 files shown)


pkra:
Thanks! PRs need to be made to `develop` (see https://github.com/mathjax/MathJax/wiki/Source-control-policies).

Otherwise, lgtm -- @dpvc?

jaredwindover:
Is there a way I can modify this pull request, or should I create a new one?

dpvc:
Unfortunately, there doesn't seem to be a way to change the target branch, so you will need to close this request and make another one.

On the other hand, I have some comments on the code that I am planning to make, so you might want to wait until after I do that.  Should be shortly.

dpvc:
Here's my thought about the code.  You are on the right track, but I think you can do it a little simpler as follows, which only requires changes to `CreatePreview()` with no need for a new function.  The new function would be

``` javascript
  mjPending: false,  // true when a typeset has been queued
    ...
  CreatePreview: function () {
    Preview.timeout = null;
    if (this.mjPending) return;
    var text = document.getElementById("MathInput").value;
    if (text === this.oldtext) return;
    if (this.mjRunning) {
      this.mjPending = true;
      MathJax.Hub.Queue(["CreatePreview",this]);
    } else {
      this.buffer.innerHTML = this.oldtext = text;
      this.mjRunning = true;
      MathJax.Hub.Queue(
        ["Typeset",MathJax.Hub,this.buffer],
        ["PreviewDone",this]
      );
    }
  },
 ...
  PreviewDone: function () {
    this.mjRunning = this.mjPending = false;
    this.SwapBuffers();
  }

```

I changed the variable name to be more in line with how this code works.  If MathJax is currently running when `CreatePreview()` is called, it queues the `CreatePreview()` for later, and sets the `mjPending` variable, which indicates that `CreatePreview()` has been queued for the future.  This will be run automatically when the current queue is completed, and we use `mjPending` to prevent additional copies of `CreatePreview()` from being queued.  The `PreviewDone()` function clears the flag, so the `CreatePreview()` that has been queued following the `PreviewDone()` will execute.  I think this more naturally reflects the logic of what needs to be done.

pkra:
Closing this in favor of #1147

PR #1147, opened 2015-03-31
Now recheck when changes made while typesetting

### LINKED ISSUE #1132: Sample handles quick changes to large content badly

jaredwindover:
Made changes as per dpvc's suggestions.
Added a flag `mjPending` to signal that the content has changed while the typesetting was happening. Setting the flag requeues `CreatePreview()`. This flag is then used for early exit instead of `mjRunning` This can close #1132.


dpvc:
The changes are good, but there is still a problem with the pull request.  You have it targeted to develop, which is great, but the problem is that it is branched from master, where it should be branched from develop.  Merges go from develop to master (creating a release), never the other direction.  You can see that yours is the wrong direction because it includes a commit at the top that is not one of the ones that you made.  Your commit history should include only your own changes, not ones due to merging a different branch into the develop branch.

So at this point there are two possible solutions:  one would be to check out develop, create a new branch from there, copy your updated sample file and commit that fresh into the new branch, and then make another new pull request for that branch having only the one commit.

The other is to rebase the current branch from master to develop using the `git rebase` command.

I suspect the first approach is easier, though it loses the earlier commit history.  In this case, that's probably fine, however.

jaredwindover:
Okay, I understand. I'll fix that and make a new PR. Thanks for your patience with this, I'm new to git and open-source contributing, and I really appreciate it.

jaredwindover:
Closing this in favor of #1148

PR #1170, opened 2015-04-22
fix typo which put a variable in global scope

christianp:
missing comma meant that SETTINGS was put in global scope

### DIFF
--- a/unpacked/extensions/toMathML.js
+++ b/unpacked/extensions/toMathML.js
@@ -28,7 +28,7 @@
 MathJax.Hub.Register.LoadHook("[MathJax]/jax/element/mml/jax.js",function () {
   var VERSION = "2.5.0";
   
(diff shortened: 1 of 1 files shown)


pkra:
Nice. (But wrong branch.)

@dpvc do we want this in hotfix-2.5.2 or just a PR to `develop`?

dpvc:
I think develop is fine.  It has been there since 2012, so I don't think it is critical to fix it as a hotfix.

pkra:
Ok. @christianp do you want to make another PR to `develop` (and based on `develop`) or do you want us to just fix this along the way?

pkra:
Accidental close. Leaving it open until we have an alternative.

christianp:
Ahhh, I keep getting the wrong branch!

PR #1184, opened 2015-05-04
UNTESTED: Include stack trace in [Math Processing Error]->menu->show as error message

cben:
If I see a [Math Processing Error], its error message is something non-informative like
`Error: Cannot read property 'firstChild' of null`.
This (or something similar) would help a lot — if not me than you after I file a better bug report :-)
[will file separately for the actual errors that prompted this]

*PLEASE TEST*.
I only tried this on Chrome and Firefox (Linux) and didn't run any test suite.
The `if(err.stack)` test sounds safe to me (and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack claims wide support) but I'm the kind of web dev who breaks stuff in prod and doesn't notice for months...

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -2327,12 +2327,13 @@ MathJax.Hub = {
   
   formatError: function (script,err) {
     //
-    //  Get the error message, URL, and line, and save it for
+    //  Get the error message, URL, line, and stack trace, and save it for
     //    reporting in the Show Math As Error menu
     //
     var message = "Error: "+err.message+"\n";
     if (err.sourceURL) {message += "\nfile: "+err.sourceURL}
     if (err.line) {message += "\nline: "+err.line}
+    if (err.stack) {message += "\nstack:\n"+err.stack}
     script.MathJax.error = MathJax.OutputJax.Error.Jax(message,script);
 
     //


dpvc:
In general, people will be using the packed version of MathJax, and so the stack trace will be of very little value (all of MathJax will be on one line, so the line number will be useless, and without that, the message doesn't tell yo very much).  

In any case, MathJax stores the error object from the last `Math Processing Error` in `MathJax.Hub.lastError`, so you can always get the stack trace from that if you need it.  Anyone who is capable of understanding the stack trace should be able to use the brewer console to access that variable, and those who aren't really have no need to see it.

So I'm inclined not to make this merge.  @pkra, do you have thoughts on this?

pkra:
> So I'm inclined not to make this merge. @pkra, do you have thoughts on this?

I'm undecided. If more people are interested, it does not seem unfathomable (well, if I ignore the IE10+ part in the browser support). I guess if we switch to uglify we could generate source maps and this becomes more viable in general. But that's a lot of ifs. More something for 3.0 perhaps?

cben:
Aha, I see.  I wondered why you didn't have it, as it's very easy — I
haven't considered that packed stacktrace is useless.

Perhaps then it'd be useful if the "show as error" message included a tip
about MathJax.Hub.lastError.stack (and using the unpacked/ version)?

PR #1209, opened 2015-07-14
Update MathZoom.js with global border-box support

CalebKester:
With a lot of sites using the global box-sizing it breaks the zoom box.  This fix will explicitly add content-box on this element to prevent a global override.

### DIFF
--- a/unpacked/extensions/MathZoom.js
+++ b/unpacked/extensions/MathZoom.js
@@ -36,6 +36,9 @@
       "#MathJax_Zoom": {
         position:"absolute", "background-color":"#F0F0F0", overflow:"auto",
         display:"block", "z-index":301, padding:".5em", border:"1px solid black", margin:0,
+        "-webkit-box-sizing":"content-box", // Android ≤ 2.3, iOS ≤ 4
+        "-moz-box-sizing":"content-box", // Firefox ≤ 28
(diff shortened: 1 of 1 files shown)


pkra:
Interesting. Could you share an example where the zoom breaks?

CalebKester:
I'm sorry, i completely forgot to put an example in.

In our internal testing we had reports if the zoom level is around 300 on large formulas it will show up like: 
chrome

It doesn't happen much but if the formula with zoom goes wider than the screen it won't trigger the overflow scroll unless it has content box on it.

firefox

pkra:
Thanks for following up on this.

@dpvc what do you think?

dpvc:
I don't have any problem with it, especially since the MathJax web site is one example of it!

dpvc:
But, the pull request needs to be made using a branch from the `develop` branch and directed at the `develop` branch, whereas yours is from a copy of `master` to `master`.  You will need to close this request and open a new one using a branch from and to `develop`.  You can cherry-pick your commit, or since it is simple, just add it in again.  If you are not familiar with git, I can do that for you.

CalebKester:
Closing pull request.  New request on develop branch: https://github.com/mathjax/MathJax/pull/1210

PR #1216, opened 2015-07-23
Since we probably can't give informative traceback, tell user how to obtain one.

cben:
Instead of #1184.
I think both points are non-obvious good leads for people who don't debug MathJax every day.


pkra:
@dpvc any thoughts?

dpvc:
I don't have a problem with it in principle, but I did notice that neither the new nor the old code are properly localized.  But that is probably a problem for us not the cben.  I'm OK with merging and fixing the localization afterward.  Do we have a CLA?

cben:
Signed the CLA now.

pkra:
@cben's CLA is confirmed -- thanks for contributing and sticking with this! We really appreciate it.

PR #1226, opened 2015-08-01
I think that this is the right URL

cheerfulstoic:


pkra:
Thanks! Nice catch. 

This is already fixed in `develop`. (We don't touch `master ` except on releases, cf. https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md#working-on-mathjax-core.)

PR #1240, opened 2015-08-17
Adds level of subclassing for clickable menu items.

zorkow:
Introduces an abstract class for menu items that are clickable, in preparation of keyboard menu navigation.

### DIFF
--- a/unpacked/extensions/MathMenu.js
+++ b/unpacked/extensions/MathMenu.js
@@ -414,6 +414,19 @@
     name: "", // the menu item's label as [id,label] pair
     node: null,
 
+    /*
+     *  Accessor method for node.
+     */
+    GetNode: function() {
+      return this.node;
+    },
+    /*
+     *  Registers the HTML node of the menu item.
+     */
+    SetNode: function(node) {
+      this.node = node;
+    },
+
     Attributes: function() {
       return {onmouseup: MENU.Mouseup,
               ondragstart: FALSE, onselectstart: FALSE, onselectend: FALSE,
@@ -424,7 +437,8 @@
       if (!this.hidden) {
         var def = this.Attributes();
         var label = this.Label(def,menu);
-        this.node = HTML.addElement(menu, "div", def, label);
+        var node = HTML.addElement(menu, "div", def, label);
+        this.SetNode(node);
       }
     },
     Name: function () {return _(this.name[0],this.name[1])},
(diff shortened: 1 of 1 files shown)


zorkow:
Mainly to play around MJs OO model. But it also makes attribute assignment cleaner, avoiding delete statements.
Use of classList API would be nice, but is that permissible (without toggle) in the light that it is IE>=10, only?

zorkow:
ptal.

unpacked/extensions/MathMenu.js
```diff
@@ -481,9 +496,29 @@
...
+      if (this.disabled) {
+        def.className += " MathJax_MenuDisabled";
+      }
+      var augdef = {onmouseover: MENU.Mouseover, onmouseout: MENU.Mouseout,
+                    onmousedown: MENU.Mousedown, role: this.role,
+                    'aria-disabled': !!this.disabled};
+      return MathJax.Hub.Insert(def, augdef);
+    }
+  });
+
```
dpvc: Actually, what I had in mind was slightly different, but along these lines.  What I was thinking was that the parent `Attributes()` method at line 430 above would be

``` javascript
Attributes: function(def) {
  return HUB.Insert({
    onmouseup: MENU.Mouseup,
    ondragstart: FALSE, onselectstart: FALSE, onselectend: FALSE,
    ontouchstart: MENU.Touchstart, ontouchend: MENU.Touchend,
    className: "MathJax_MenuItem", menuItem: this
  },def);
},    
```

and then the one here could be

``` javascript
Attributes: function() {
  var def = this.SUPER(arguments).Attributes.call(this,{
    onmouseover: MENU.Mouseover, onmouseout: MENU.Mouseout,
    onmousedown: MENU.Mousedown, role: this.role, 'aria-disabled': !!this.disabled
  });
  if (this.disabled) def.className += " MathJax_MenuDisabled";
  return def;
}
```
Here, `HUB.Insert(obj,def)` will do nothing if `def` is empty or `null`, so the calls to `this.Attributes()` will work as before, but the overridden `Attributes` here will apply the additional attributes.  Of course, it certainly works as is, and I will merge if you prefer to leave it how it is now.

zorkow:
I see. That is a philosophical difference. I always believe in a class providing the minimum logic for its purpose and subclasses adding to it for their own needs. This way one does not have to anticipate in a class what potential extensions might need.
One could argue of course that Menu.Entry is effectively an abstract class and hence its job is to add a certain default to its concrete subclasses. I am fine with that, and indeed that is often the purpose of abstract classes.

But my real problem with this particular example is that it creates an interface mismatch: 
The Entry class has Attributes(x) in its interface, while (at least one of) its subclasses have Attributes().
One reason why I like clean interface specifications. [Of course in an ideal world Attributes would be protected and hence not part of the class interface.]

Anyway, I am happy to go with your, more concise solution; but we might consider some of these issues for 3.0.

zorkow:
Ready to merge.

dpvc:
> my real problem with this particular example is that it creates an interface mismatch

Yes, I was not happy about that, either, and had considered making the one in `ENTRY` also having a parameter;  Indeed, that was my original plan, as I had through some of its subclasses used that, but it turned out that after adding the `role` property, that was no longer the case.  Adding the `def` parameter there would have involved another call to `HUB.Insert()` that was never used, so I left it out.  We can certainly add that back in if that makes you happier (I was on the fence about it).

Alternatively, since there are no additional attributes being passed in (as I thought there were going to be), we could go back to your original.  You can decide.  If you want to make the change, do so, otherwise, we can merge it as is.

zorkow:
I'll go ahead and merge for now as it will be a very ephemeral state anyway.

I have now the main refactoring of the Menu to work with keyboard and focusing ready.
Although there are still a couple of issues remaining, I'd like to push it out so you can have look at it. 
And it is currently based off of this branch, so any further change here will annoy me further up.

dpvc:
Sounds good.

Issue #939, opened 2014-10-21
Make MathJax contextual menu properly accessible


jcsteh:
The MathJax contextual menu inserts clickable divs at the end of the document, but it does not focus them in any way or expose semantic information. For screen reader users (and perhaps users of other assistive technology products), these are difficult to locate. Most users wouldn't even realise this menu had appeared, let alone know how to use it. This can all be fixed with ARIA and keyboard handling.

Things that need to be done:
* The menu items need to be focused. Either set tabindex="-1" for all menu items and focus the item or set tabindex="-1" on the root of the menu, focus that and set aria-activedescendant on it to indicate the active item.
* Menus need to be given an ARIA role of menu. Menu items need to be given appropriate menu item roles (menuitem, menuitemcheckbox, menuitemradio).
* There needs to be an appropriate hierarchical relationship between submenu items and their submenu. Either the DOM positioning needs to be changed or aria-owns needs to be used.
* Appropriate keyboard navigation needs to be implemented; e.g. up/down arrows to move between items, left/right arrows to enter/exit submenus, enter to activate an item.

/cc @pkra

pkra:
Just some notes from other discussions.

* we probably need to disable outlines caused by adding equations to the tabindex (since browser do not honor clipping, making the outlines look extremely messy/irritating).
* so we need to provide a visual cue (maybe along the lines of the unfinished discvoerability UI).

pkra:
Just to update this thread. 

From #1242. 

## TODO

* [ ] I've introduced an abstract class with default methods of keyboard navigatable objects as a superclass of menu and item but reattached it as a field to MathJax.Menu, which is slightly awkward.
* [ ] Opening menu by keyevent (Space) enables left right navigation through Math element similar to a menu bar. Not sure if we want that. Maybe with a modifier key?
* [ ] On Chrome focusing seems to also cause scroll events if the menu is too large, which can cause awkward interactions with the mouse pointer. Not sure yet how to avoid this.
* [ ] I currently use the standard right-open/left-close paradigm for submenus. This can be visually confusing when the submenu opens to the left, but seems to be natural given that the pointer is announced as right pointer.
* [ ] I still need to deal with the case of rtl menus.
* [ ] Focus on popups is not yet correct. This is particularly not straight forward for windows showing code.
* [ ] Aria elements might still need refinement.

pkra:
I believe we can close this as fixed by https://github.com/mathjax/MathJax/pull/1242

[...]

PR #1242, opened 2015-08-21
Integrate keyboard events

zorkow:
Integrated keyboard navigation and aria roles into menu navigation. 
There are still a few issues and TODOs, which I will go over in a separate comment.
(Sorry, by force of habit I did a whitespace cleanup, which might make things harder to review...)


pkra:
Great progress! Is there a todo list somewhere? I'd like to cross post to #939.

FYI, keyboard navigation on top-level elements with no submenu lets me cycle through equations. Is that intentional or TODO?

pkra:
With "navigation" I meant hitting left/right keys will move me between equations.

zorkow:
Here's my snag list for the PR. Some should be sorted out before merging, some can probably wait. 

-- I've introduced an abstract class with default methods of keyboard navigatable objects as a superclass of menu and item but reattached it as a field to MathJax.Menu, which is slightly awkward.
-- Opening menu by keyevent (Space) enables left right navigation through Math element similar to a menu bar. Not sure if we want that. Maybe with a modifier key?

-- On Chrome focusing seems to also cause scroll events if the menu is too large, which can cause awkward interactions with the mouse pointer. Not sure yet how to avoid this.
-- I currently use the standard right-open/left-close paradigm for submenus. This can be visually confusing when the submenu opens to the left, but seems to be natural given that the pointer is announced as right pointer.
-- I still need to deal with the case of rtl menus.
-- Focus on popups is not yet correct. This is particularly not straight forward for windows showing code.
-- Aria elements might still need refinement.

-- ... and do we already have tests for the menu?

pkra:
Awesome, thanks!

Tests are at https://github.com/mathjax/MathJax-test/tree/master/testsuite/UI

dpvc:
> Sorry, by force of habit I did a whitespace cleanup, which might make things harder to review...

No problem, I can deal with it.  I've taken a quick look at the code and it looks good.  I do want to try it out, but won't get to that immediately.  In any case, I don't see anything but a few small items.  The only real issue is that the math on the page can change at any time (dynamic pages add and remove math via javascript), so the cached results of `GetJaxes()` can become invalid.  You could listen for "New Math" signals and either invalidate the cache or add the new jax to it.  Also, only the HTML-CSS output has class name equal to `MathJax`.  Others, use `MathJax_SVG`, for example.  We could add another class that is common to all the output formats that identifies the element with the menu events.  That might make things easier.

pkra:
@zorkow I prettified your todo list comment. Hope that's ok.

zorkow:
@pkra Thanks. I've also added one more point.

zorkow:
> The only real issue is that the math on the page can change at any time (dynamic pages add and
 > remove math via javascript), so the cached results of GetJaxes() can become invalid. 
> You could listen for "New Math" signals and either invalidate the cache or add the new jax to it.

Right, I did not think about this, when doing the caching. In general, caching is probably something we should get rid of for this release entirely, as it does not seem to be consistent with the practice in 2.x. We can have a rethink for 3.0.

Btw., can nodes change while a menu is being displayed? Or rather the other way around: Can a menu be displayed while a node changes? Which part is inserted first, the Frame where the event handler sits?

dpvc:
> In general, caching is probably something we should get rid of for this release entirely, as it does not seem to be consistent with the practice in 2.x.

Probably best to leave it out for now.  In addition to new math showing up, it is also possible for old math to disappear without warning (and there are no signals for this, either).  For example, if someone sets `innerHTML` on a container element that contains math.

>  We can have a rethink for 3.0.

Sure.

> can nodes change while a menu is being displayed?

Yes.  Javascript can run while the menu is open and you are waiting for user activity.  So new math can show up, old can be removed, and the menu itself could be removed (by javascript setting `innerHTML` on a container that is a parent of the menu node).  Of course, the DOM nodes will not be removed from memory until you no longer hold pointers to them, so you can safely manipulate them, for the most part, but they will not be on screen any more.

> Or rather the other way around: Can a menu be displayed while a node changes?

Yes, I suppose it could.  For example, javascript could cause the math to be re-rendered (e.g., for a change in content), and that would cause the current output to be removed and recreated.  That includes the Frame container.

Another possibility is with the automatically collapsing equations from the Sloan work.  If the menu is showing and the window is resized and an equation collapses or expands, currently that causes a re-render, which would remove the Frame element.

> Which part is inserted first, the Frame where the event handler sits?

It is output jax dependent, but the HTML-CSS output first add the Frame element (with the event handles in place) and then adds the HTML-CS rendering inside it (so that it can measure the results if needed while it is processing).  Normally, the creation of the HTML will be atomic, so no other javascript can run while that is happening, but that is not always the case.  If MathJax needs to load an external file (e.g., load a font, or a font data file, or an extension, or some other such thing), then MathJax will pause and other javascript can run while an equation is only partly created.

But in these cases, the frame will be set to `display:none`, so you shouldn't be able to select it.  If you want to check that programmatically, there is a class added to the frame that could be used to detect that.

unpacked/extensions/MathEvents.js
```diff
@@ -144,6 +158,19 @@
     },
 
     //
+    // Keydown event handler. Should only fire on Space key.
+    //
+    // TODO: (sorge) Fit this into the Handler function.
+    //
+    Keydown: function (event, math) {
+      var jax = OUTPUT[this.jaxID];
```
dpvc: The `jax` variable doesn't seem to be used anywhere in this routine.

unpacked/extensions/MathMenu.js
```diff
@@ -400,19 +472,105 @@
...
+    node: null,   // The node the menu was activated on.
+    active: null,   // The currently focused item. There can only be one!
+    posted: false,  // Is a menu open?
+
+    GetJaxs: function() {
+      var nodes = document.getElementsByClassName('MathJax');
+      for (var i = 0, node; node = nodes[i]; i++) {
+        MENU.jaxs.push(node);
+      }
+    },
```
dpvc: Just a reminder:  we've already discussed the issue of caching the math nodes.

unpacked/extensions/MathMenu.js
```diff
@@ -400,19 +472,105 @@
...
+     */
+    jaxs: [],    // List of all MathJax nodes.
+    hasJaxs: false,  // Flag to indicate if the MathJax node list has already
+                     // been computed.
+    node: null,   // The node the menu was activated on.
+    active: null,   // The currently focused item. There can only be one!
+    posted: false,  // Is a menu open?
+
+    GetJaxs: function() {
+      var nodes = document.getElementsByClassName('MathJax');
```
dpvc: Note that `getElementByClassName()` is not available in IE8 and below.  Also, the `MathJax` is only used by the HTML-CSS output, so this won't find the output produced by other jax.  The SVG jax uses `MathJax_SVG`, and the NativeMML output use `MathJax_NativeMML` (though it is on the parent of the element that you probably want), while the CommonHTML jax doesn't use a class at all.

I think what you want to do is something like

``` javascript
var jaxs = MathJax.Hub.getAllJax();
for (var i = 0, jax; jax = jaxs[i]; i++) {
}
```

dpvc: [Sorry, accidentally posted too soon.]

Note that `getElementByClassName()` is not available in IE8 and below.  Also, the `MathJax` is only used by the HTML-CSS output, so this won't find the output produced by other jax.  The SVG jax uses `MathJax_SVG`, and the NativeMML output use `MathJax_NativeMML` (though it is on the parent of the element that you probably want), while the CommonHTML jax doesn't use a class at all.

I think what you want to do is something like

``` javascript
var jax = MathJax.Hub.getAllJax();
for (var i = 0; jax[i]; i++) {
  jax[i] = document.getElementById(jax[i].inputID+"-Frame");
  if (!jax[i].isMathJax) jax[i] = jax[i].firstChild;
}
```

Of course, if you are only using this to move to the next or previous math node, it might not be necessary to look up every element each time.  Just use `getAllJax()`, find your jax in that list, move to the next or previous one, and then look up the Frame element.

unpacked/extensions/MathMenu.js
```diff
@@ -400,19 +472,105 @@
...
+      MENU.posted = false;
+    },
+    //TODO: A toggle focus method on the top level would avoid having to
+    //tabIndex all the Jaxs.
+    Move: function(event, menu, move) {
+      var len = MENU.jaxs.length;
+      if (len === 0) {
+        return;
+      }
+      var next = MENU.jaxs[MENU.Mod(move(MENU.jaxs.indexOf(MENU.node)), len)];
```
dpvc: Note that IE8 and below does not support `indexOf()`, so this will break in IE8.

unpacked/jax/output/HTML-CSS/jax.js
```diff
@@ -569,8 +569,9 @@
         span = div = this.Element("span",{
 	  className:"MathJax", id:jax.inputID+"-Frame", isMathJax:true, jaxID:this.id,
           oncontextmenu:EVENT.Menu, onmousedown: EVENT.Mousedown,
-          onmouseover:EVENT.Mouseover, onmouseout:EVENT.Mouseout, onmousemove:EVENT.Mousemove,
-	  onclick:EVENT.Click, ondblclick:EVENT.DblClick
+          onmouseover:EVENT.Mouseover, onmouseout:EVENT.Mouseout,
+          onmousemove:EVENT.Mousemove, onclick:EVENT.Click,
+          ondblclick:EVENT.DblClick, onkeydown: EVENT.Keydown, tabIndex: "0"
```
dpvc: There will need to be similar additions to the SVG, NativeMML, and CommonHTML output as well.

unpacked/extensions/MathEvents.js
zorkow: Done.

unpacked/extensions/MathMenu.js
zorkow: There is no more caching.

unpacked/extensions/MathMenu.js
zorkow: Done.

unpacked/extensions/MathMenu.js
zorkow: Added to prototype if necessary.
We might want to do this somewhere globally as the function is also used in 
./jax/output/SVG/jax.js:122:
./jax/input/AsciiMath/jax.js:832:
./jax/input/AsciiMath/jax.js:833:
./jax/input/AsciiMath/jax.js:834:
./jax/input/AsciiMath/jax.js:835:
./jax/input/AsciiMath/jax.js:924:
./jax/input/AsciiMath/jax.js:925:
./jax/input/AsciiMath/jax.js:926:
./jax/input/AsciiMath/jax.js:1213:
./jax/input/AsciiMath/jax.js:1266:
./jax/input/AsciiMath/jax.js:1267:
./extensions/MathML/content-mathml.js:96:

unpacked/jax/output/HTML-CSS/jax.js
zorkow: Done.

zorkow:
I've done a major revision which mainly:
- removes caching
- deals with rtl/ltr correctly
- cleans focus vs activation code
- deals correctly with most popup elements

unpacked/extensions/MathMenu.js
dpvc: Note that the AsciiMath references are to the String object's `indexOf()`, not the Array object's `indexOf()`.  IE8 and below do support the String version, but not the Array one.  The SVG jax does use it, but since SVG is not supported in IE8 and below anyway, that's not really a problem (it would fail in any case), so I didn't worry about using it there.

As for changing the Array `prototype`, we have a policy of not doing that, and I do feel fairly strongly about that.  Because MathJax is a library that authors use in conjunction with other libraries and in pages that we don't control, we should not make any more global changes to the javascript environment than we have to.  Modifying base object prototypes is a kind of change that can have unexpected consequences, and people who use MathJax would not expect it to be altering the functionality of the Array object.  Since this can be handled easily in other ways, I would not want to break this long-standing principle for this situation.

My recommendation would be to do something like

``` javascript
var IndexOf = (Array.prototype.indexOf ?
  function (A,item,start) {return A.indexOf(item,start)} :
  function (A,item,start) {
    for (var i = (start || 0), j = A.length; i < j; i++) {if (item === A[i]) return i}
    return -1;
  });
```

earlier on, and then use `IndexOf(MENU.jaxs,MENU.node)` (in the case of the line above).

We could make a `MathJax.Util` object that contains this kind of service routine (`MathJax.Hub.Insert` could be moved there, as well as your `Mod` function) if you think that would help.

unpacked/extensions/MathMenu.js
```diff
@@ -400,32 +472,98 @@
...
+    AllNodes: function() {
+      var jaxs = MathJax.Hub.getAllJax();
+      var nodes = [];
+      for (var i = 0, jax; jax = jaxs[i]; i++) {
+        nodes.push(MENU.GetNode(jax));
+      }
+      return nodes;
+    },
+    ActiveNode: function() {
+      return document.activeElement;
```
dpvc: I'm a little concerned about `document.activeElement`, since that can change while the menu is open.  (E.g., some other javascript focuses another element.)

I'm not against caching the active element for the menus (the change of memory loss is very low; I think it is only if the page closes while the menu is posted).

Alternatively, you might need to have an event for change of focus?  Or you could check if the `activeElement` is actually the one you marked with the active class.  Or you could even attach a pointer to the active element to the main menu DOM element and grab it from there.

In any case, I think this is fragile as it stands.

unpacked/extensions/MathMenu.js
```diff
@@ -503,20 +653,84 @@
...
+      this.MoveHorizontal(event, item, MENU.Left, this.isRTL());
+    },
+    MoveHorizontal: function(event, item, move, rtl) {
+      if (ITEM.GetMenuNode(item).menuItem === MENU.menu) {
+        move(event, item);
+      }
+      if (rtl) { return; }
+      this.Deactivate(item);
+      var sibling = item.parentNode.previousSibling;
+      var actives = sibling.getElementsByClassName('MathJax_MenuActive');
```
dpvc: The `getElementsByClassName()` method is another one of those that isn't supported in IE8 and below.  One could scan the children for active elements by hand, I suppose.

unpacked/extensions/MathMenu.js
```diff
@@ -1240,19 +1498,31 @@
...
 
   CALLBACK.Queue(
     HUB.Register.StartupHook("End Config",{}), // wait until config is complete
     ["Styles",AJAX,CONFIG.styles],
     ["Post",HUB.Startup.signal,"MathMenu Ready"],
-    ["loadComplete",AJAX,"[MathJax]/extensions/MathMenu.js"]
+    ["loadComplete",AJAX,"[MathJax]/extensions/MathMenu.js"],
+    function() {
+      if (!Array.prototype.indexOf) {
+        Array.prototype.indexOf = function(item, start) {
```
dpvc: See my [earlier note](https://github.com/mathjax/MathJax/pull/1242#discussion_r38262360) about modifying modifying object prototypes.

unpacked/extensions/MathEvents.js
```diff
@@ -117,19 +131,22 @@
...
     //  Call the output jax's event handler or the zoom handler
     //
     Handler: function (event,type,math) {
+      console.log(type);
       if (AJAX.loadingMathMenu) {return EVENT.False(event)}
       var jax = OUTPUT[math.jaxID];
+      console.log(jax);
       if (!event) {event = window.event}
       event.isContextMenu = (type === "ContextMenu");
+      console.log(jax[type]);
```
dpvc: These `console.log()` commands need to be removed.

dpvc:
OK, I've had a chance to experiment with the changes, and overall, I like what you have done.  There are a few issues that I've found, and some questions about how to handle pointer versus keyboard events.

Here are the issues I spotted, in no particular order:

*  When the initial menu opens via keyboard, no entries are selected (which is fine), and you use down/up keys to select them (also fine).  When a submenu is selected, the submenu opens with nothing selected (also fine).  Right arrow selects the first entry in the submenu (as expected).  Subsequent submenus can be selected in the same way.  When an entry in a submenu is selected, the left arrow selects the entry in the parent menu and the submenu remains open but has nothing selected.  All that is fine.  But if the main menu has a submenu entry selected (so the submenu is open but nothing in it is selected), I would think the left arrow would unselect the entry in the main menu (so the main menu is open with nothing active) and close the submenu.  Instead, left arrow moves you to the previous math element and opens its menu.  I was regularly caught by this, as it does not work as I would expect.

* When an entry in the main menu is active and you use left or right arrow, this closes the menu and opens the menu for the net or previous math element.  This seems strange to me.  Is that really the expected behavior?

* Would it make sense to have space close the menu if nothing is active in the menu?  I know escape closes it, but for some reason my natural instinct was to hit space again if I opened a menu and wanted to close it without selecting anything.

* Volker already pointed out that even when a submenu opens to the left, you need to use the right arrow to move into it.  That does take some getting used to.  It seems more natural to use the left arrow.  Would it be a mistake to implement that?

* If the page has been scrolled, then opening a menu via keyboard has the menu open at the upper left, not next to the math.  After that, using right and left arrows to move to other math causes the math to open in the wrong place on the page and it is not clear what math the menu belongs to.

* This brings up an interface question: should we highlight the math that has its menu open so that it is clear which one it goes to?  Right now it is sometimes hard to tell.

* Disabled submenu items (like the Annotations item in the Show Math As submenu) can be selected by keyboard (but not by mouse).  Should that be allowed?  Note that left and right arrows don't work from such an item, even though it is focused.

* When a menu is open, keyboard characters that *aren't* used for menu navigation get lost (e.g., you can't reload the page via keyboard when a menu is open).  Should unused keyboard events be passed through rather than ignored?

* Currently, the zoom function is not available from the keyboard.  Should there be a way to zoom when a math element is selected?  (And if so, should it be controlled by the "Zoom Trigger" menu in some way?)

Some issues about focus:

*  We probably want to style the HTML-CSS output since otherwise the focus outline is too large or (in some browsers) strangely shaped.  One could add CSS to turn off the usual outline and use a border instead, though it would be difficult to maintain the same look as the native focus outline since it varies from browser to browser.  Perhaps that is not such a big deal.  A version compatible with Firefox would be to add
    
        ".MathJax:focus": {display:"inline-block", outline:"none", border:"1px dotted", margin:"-1px"},
    
    to the `styles` property of the HTML-CSS default configuration.  The default WebKit focus outline has transparency or and fades at the edges, and I haven't been able to reproduce that yet.  I haven't checked what IE does.

* The menus and dialog boxes show the focus outline even when they weren't opened via keyboard.  I would expect that if I used the mouse to open the menu or dialog box, it would not show a focus outline (i.e., would not gain the focus).  Can the focus control be conditional on the type of event that initiated the change?  I find this most disturbing for the pop-up windows, as I am not used to seeing focus outlines as a mouse user.

* Similarly, when the menu closes after actions like "Scale All Math", the math element becomes focused even if I haven't been using the keyboard to manage the menu.  This is unexpected.

* Opening the localization submenu when the main menu is near the bottom of the window can cause the page to scroll (presumably due to the focus for the submenu), though this seems to be browser dependent.  This may not be so bad for keyboard users, but for mouse users this can be a problem.  It also is different from the current situation, where the page stays in place and the menu moves off the bottom of the window.  If you don't change focus for mouse menu interactions, that would probably take care of it.  Otherwise, I'm not sure anything can be done about it.

* Do we want to style the focus CSS for the menu entries (since we already have a visual indication of the selected menu item)?  Currently, the focus outline can be distracting.

zorkow:
Thanks for the list. I am currently working on some issues (e.g. involuntary mouse/keyboard interaction)
I'll reply to some of your points first:


* When the initial menu opens via keyboard, no entries are selected (which is fine), and you use down/up keys to select them (also fine). When a submenu is selected, the submenu opens with nothing selected (also fine). Right arrow selects the first entry in the submenu (as expected). Subsequent submenus can be selected in the same way. When an entry in a submenu is selected, the left arrow selects the entry in the parent menu and the submenu remains open but has nothing selected. All that is fine. But if the main menu has a submenu entry selected (so the submenu is open but nothing in it is selected), I would think the left arrow would unselect the entry in the main menu (so the main menu is open with nothing active) and close the submenu. Instead, left arrow moves you to the previous math element and opens its menu. I was regularly caught by this, as it does not work as I would expect.

This reflects regular keyboard driven pull down menu behaviour as
in standard applications (e.g. in Office applications).
* Space/Enter opens commands/submenus. Also closes the latter.
* Esc leaves menu from anywhere.
* Once a menu items become active the never become inactive
* Right moves into submenu's. Left out of submenus.
* Left/right moves to the next menu point. In our case I've used
  the math element as menu point. As I said in the snag list,
  that is just something I implemented, but we might not want
  that or hide it.

* When an entry in the main menu is active and you use left or right arrow, this closes the menu and opens the menu for the net or previous math element. This seems strange to me. Is that really the expected behavior?


* Would it make sense to have space close the menu if nothing is active in the menu? I know escape closes it, but for some reason my natural instinct was to hit space again if I opened a menu and wanted to close it without selecting anything.

We could do that. Would somewhat be in line with other keyboard
driven menu, where for example Alt-F opens file menu and also
closes it if pressed again.

zorkow:
* Volker already pointed out that even when a submenu opens to the left, you need to use the right arrow to move into it. That does take some getting used to. It seems more natural to use the left arrow. Would it be a mistake to imp

I think that would be confusing. Imagine you can not see the
menu. Your screen reader tells you right arrow, you open the
submenu, but it visually appears on the left. Then right arrow
will not get you into the submenu.
Going left currently is reserved for RTL menus.

zorkow:
* If the page has been scrolled, then opening a menu via keyboard has the menu open at the upper left, not next to the math. After that, using right and left arrows to move to other math causes the math to open in the wrong place on the page and it is not clear what math the menu belongs to.

That is this indeed a real pain. The real problem is that
getBoundingClientRect is a function relative to the viewport and
somehow is updated later (e.g. after the scrolling). Once it
actually gives you a negative argument, this is the behaviour you
see. I have experimented with a couple of different ways to get
around that, but was so far unsuccessful.

zorkow:
* This brings up an interface question: should we highlight the math that has its menu open so that it is clear which one it goes to? Right now it is sometimes hard to tell.

True, but again the question is: Since the keyboard driven menu
is mainly aimed at VI users, visual highlighting might not help
them.


* Disabled submenu items (like the Annotations item in the Show Math As submenu) can be selected by keyboard (but not by mouse). Should that be allowed? Note that left and right arrows don't work from such an item, even though it is focused.

That's WCAG guidelines. If a submenu item is preceivable it has
to be possible to voice it. Focusing on it after a key event is
the way to consistently achieve this.  This is particularly
important for labels like "Trigger requires..."

* When a menu is open, keyboard characters that aren't used for menu navigation get lost (e.g., you can't reload the page via keyboard when a menu is open). Should unused keyboard events be passed through rather than ignored?

Yes, we should do that.

zorkow:
* Opening the localization submenu when the main menu is near the bottom of the window can cause the page to scroll (presumably due to the focus for the submenu), though this seems to be browser dependent. This may not be so bad for keyboard users, but for mouse users this can be a problem. It also is different from the current situation, where the page stays in place and the menu moves off the bottom of the window. If you don't change focus for mouse menu interactions, that would probably take care of it. Otherwise, I'm not sure anything can be done about it.

It's even worse, even if you use keyboard and the mouse pointer 
is at the wrong position, one can get interferences. I am currently
experimenting with ways to avoid that.
I think the real problem here is, if you intersperse mouse and
keyboard interaction. One would expect a consistent
behaviour. I.e., focus on item n, mouse goes to item m, next
keyboard interaction goes to either m-1 or m+1.  Also the only
reliable way to pass the active element to an AT is by focusing
it.

unpacked/extensions/MathMenu.js
```diff
@@ -193,28 +193,79 @@
       },
       ".MathJax_MenuClose:hover span": {
         "background-color":"#CCC!important"
+      },
+      ".MathJax:focus": {
+        display:"inline-block",
+        outline:"none",
+        border:"1px dotted",
+        margin:"-1px"
```
dpvc: Here is code that makes things better match the default WebKit focus outline (at least on Mac OS).  I still need to check on Windows, and what IE does by default.

``` css
      ".MathJax:focus": (
        (MathJax.Hub.Browser.isSafari || MathJax.Hub.Browser.isChrome) ? {
          display:"inline-block",
          outline:"none",
          margin:"-3px",
          padding:"3px",
          "-webkit-box-shadow": "0px 0px 5px #345, inset 0px 0px 5px #345",
          "box-shadow": "0px 0px 5px #345, inset 0px 0px 5px #345"
        } : {
          display:"inline-block",
          outline:"none",
          border:"1px dotted",
          margin:"-1px"
        })
```

But this should really go in the `HTML-CSS/jax.js` since the `.MathJax` class is part of that output.

zorkow: Thanks. Integrated and moved.

pkra:
Keyboard navigation does not seem to work with the Fast/PreviewHTML output.

dpvc:
I was going to point that out, but was in the middle of renaming the FastHTML as PreviewHTML, and didn't want to cause conflicts if changes were made in this branch in the meantime.

dpvc:
To be more complete, the `tabindex:0` needs to be added to PreviewHTML as it was to the other output jax.

We are probably going to get merge conflicts anyway because of the renaming of CommonHTML to FastHTML.

zorkow:
I'll have a look. Should not be big.

The last commit adds the closing of submenus on left key as we have discussed today.
Apart from the occassional positioning problems for menus, this should then be all.

zorkow:
I've added the keyboard menu to FastHTML. 
In display mode the focus elements takes the whole line, which makes the menu pop up on the outer right. I'll have another look at the clientrectangle method.

zorkow:
Menus do now consistently open in the centroid of the maths element.
There is still one minor glitch: when jumping from the very top to the very bottom of the page (or vice versa) the viewport does not always change. I.e., while the menu opens on the correct element, it might not be visible without explicit scrolling.

But apart from this, I think we are pretty much done with this PR.

pkra:
:+1:

zorkow:
I've made the focus sequence contiguous, which fixes the final problem of not scrolling to first/last element when moving between maths elements.
So from my point of view the PR is finished and could be merged.

dpvc:
> So from my point of view the PR is finished and could be merged.

Great!  Thanks for all your work on this.  I want to run through it again now that you have made the changes, but I won't be able to do that until tomorrow, most likely.  But I think you did a fantastic job.

zorkow:
NP.
While making the code uniform to the current style, I had thoughts on some of the style things we should discuss for 3.0. Where should I start a thread on that?

pkra:
> Where should I start a thread on that?

This repository's wiki?

dpvc:
This page says that the merge has conflicts, but I'm able to merge by hand without trouble, so I'm not sure what is up with that.  @zorkow, do you mind if I merge by hand and close this pull request?

zorkow:
No conflicts, but I had to merge the latest version of the develop branch.

pkra:
:+1:

pkra:
:mega: :tada:

Issue #1250, opened 2015-08-31
Review mml3.js updates for v2.6


pkra:
Just a tracking issue to remember to check for udpates.

pkra:
@dpvc I wanted to add 

```javascript
MathJax.Hub.Config({
  "HTML-CSS": {
    styles: {
      ".MathJax_Display span": {
        "direction": "ltr !important",
        "display": "inline-block !important"
      }
    }
  }
});
```

to the `mml3.js` extension but it doesn't seem to work. What am I'm doing wrong?

dpvc:
By "not working" do you mean the styles are not being inserted into the page, or do you mean that the result of adding them isn't what you expect it to be?

pkra:
They didn't seem to be inserted into the page.

dpvc:
The configuration for HTML-CSS probably has already completed by the time the extension is loaded, and so the styles are already in place before you add these into the HTML-CSS configuration.  You may just have to add the style sheet yourself via a call to `MathJax.Ajax.Styles()`.

In any case, I think you want `".MathJax span"` not `".MathJax_Display span"` otherwise you will miss the inline math.

pkra:
Ah, ok. I had followed the wrong example then. That works.

pkra:
=> Merged.

Issue #1252, opened 2015-09-02
mml3js xslt transforms failing on MS Edge [was: Readium on MS Edge:  Unable to get property 'documentElement' ]


pkra:
Via https://github.com/readium/readium-js-viewer/issues/394

On http://development.readium.divshot.io/?epub=epub_content%2Flinear-algebra with Edge, MathJax throws

`Error: Unable to get property 'documentElement' of undefined or null reference`

Probably caused by mml3.js, i.e., https://github.com/mathjax/MathJax/blob/master/unpacked/extensions/MathML/mml3.js#L92-L107 -- that ActiveXObject seems dubious (Edge kicked those kinds of things out, no?)

dpvc:
I would expect Edge to implement `XSLTProcessor`, so the lines you indicate would not be reached.  In any case, they don't contain the `documentElement` call that is causing the error.  That seems to be the section below that, intended for IE8 and below.  If you use the MathJax menu to get the About MathJax box, what browser and version does it think you are using?

If the error is actually from line 116 (you can view `MathJax.Hub.lastError.stack` in the console to see more information about the call stack), then the problem is that the parsing of the original MathML is failing, and the doc at line 58 is null.  There should probably be a line

    if (doc == null) MATHML.Error(["ErrorParsingMathML","Error parsing MathML"]);

after line 58 to check for that.

pkra:
`MathJax.Hub.lastError.stack`points to https://github.com/mathjax/MathJax/blob/master/unpacked/extensions/MathML/mml3.js#L70

```
MathJax.Hub.lastError.stack
"TypeError: Unable to get property 'documentElement' of undefined or null reference
   at Anonymous function (http://192.168.0.161:8080/unpacked/extensions/MathML/mml3.js?rev=2.5.3:70:7)
   at CALLBACK.prototype.execute (http://192.168.0.161:8080/unpacked/MathJax.js:236:9)
   at cb (http://192.168.0.161:8080/unpacked/MathJax.js:218:27)
   at Execute (http://192.168.0.161:8080/unpacked/MathJax.js:419:9)
   at Translate (http://192.168.0.161:8080/unpacked/jax/input/MathML/jax.js?rev=2.5.3:263:7)
   at MathJax.Hub.processInput (http://192.168.0.161:8080/unpacked/MathJax.js:2179:11)
   at CALLBACK.prototype.execute (http://192.168.0.161:8080/unpacked/MathJax.js:236:9)
[...]

PR #1257, opened 2015-09-04
Improve RTL support in HTML-CSS

pkra:
Add CSS to prevent browser interference

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -3090,6 +3090,7 @@ MathJax.Hub.Startup = {
     isMac:       (navigator.platform.substr(0,3) === "Mac"),
     isPC:        (navigator.platform.substr(0,3) === "Win"),
     isMSIE:      ("ActiveXObject" in window && "clipboardData" in window),
+    isMsEdge:    (!!AGENT.match(/Edge\//)),
     isFirefox:   (!!AGENT.match(/Gecko\//) && !AGENT.match(/like Gecko/)),
     isSafari:    (!!AGENT.match(/ (Apple)?WebKit\//) && !AGENT.match(/ like iPhone /) &&
                      (!window.chrome || window.chrome.loadTimes == null)),
--- a/unpacked/extensions/MathML/mml3.js
+++ b/unpacked/extensions/MathML/mml3.js
@@ -80,7 +80,7 @@ MathJax.Hub.Register.StartupHook("MathML Jax Ready",function () {
    */
   var BROWSER = MathJax.Hub.Browser;
   var exslt = '';
-  if (BROWSER.isMsEdge) {
+  if (BROWSER.isMsEdge || BROWSER.isMSIE) {
     exslt = 'urn:schemas-microsoft-com:xslt'
   } else {
     exslt = 'http://exslt.org/common';
--- a/unpacked/extensions/MathML/mml3.js
+++ b/unpacked/extensions/MathML/mml3.js
@@ -72,17 +72,661 @@ MathJax.Hub.Register.StartupHook("MathML Jax Ready",function () {
   });
 
   /*
-   *  The following is taken from mml3mj.xsl 
-   *  (https://web-xslt.googlecode.com/svn/trunk/ctop/mml3mj.xsl)
+   *  The following is derived from mml3mj.xsl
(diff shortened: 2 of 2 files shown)


pkra:
Fixes #1250

pkra:
Marking this WIP in the hopes of doing something about #1252. But perhaps that's too optimistic.

pkra:
I would like to add the fix for Edge before merging this. But that can wait until the beta is out.

pkra:
Added a fix for #1252, including browser detection for MS Edge in MathJax.js.

Ready for review.

pkra:
Hold off on that review. A more extensive browser test revealed additional issues on IE.

pkra:
Adding the Edge hack for MSIE might be sufficient for the time being. 

To be precise: using http://codepen.io/pkra/pen/MaYeBZ I see

* IE11 and IE10 
  * all clear 
  * without Edge hack: 1,6,8 get processing errors
* IE9: 
  * failed: 3,4, 7 get processing errors, 6 (stray p node in first part)
  * without Edge hack: 1, 3, 4, 6, 7, 8 get processing errors
* IE8: 
  * failed: 1 (layout: full underline), 6 (processing error in first part), 8 (layout: full underlines) 
  * without Edge hack: 1, 6, 8 get processing errors
* IE7: 
  * 1 (layout: full underline), 6 (stray p node in first part), 8 (layout: full underlines) 
  * without Edge hack: 1, 6, 8 get processing errors

pkra:
I think it's ready for review as the remaining issues seem to be upstream issues. (I suspect the improvements from 2.5 never worked on older IE :-( )

unpacked/MathJax.js
```diff
@@ -3090,6 +3090,7 @@ MathJax.Hub.Startup = {
     isMac:       (navigator.platform.substr(0,3) === "Mac"),
     isPC:        (navigator.platform.substr(0,3) === "Win"),
     isMSIE:      ("ActiveXObject" in window && "clipboardData" in window),
+    isMsEdge:    (!!AGENT.match(/Edge\//)),
```
dpvc: I thought you meant that you were going to use `!!AGENT.match(/Edge\//)` in `mml3.js` itself, not modify `MathJax.js`.  The latter requires additional work:  you would need to handle the version number later in the code, and you would also need to make sure that `isSafari` isn't also set (which it still will be).

unpacked/extensions/MathML/mml3.js
```diff
@@ -121,6 +765,9 @@ MathJax.Hub.Register.StartupHook("MathML Jax Ready",function () {
     // No XSLT support. Do not change the <math> content.
     MATHML.mml3XSLT = null;
   }
+  
+  // Tweak CSS to avoid some browsers rearranging HTML output
+  MathJax.Ajax.Styles(".MathJax span { direction: ltr !important; display: inline-block !important;}");
```
dpvc: This is probably over-aggressive, and may cause unforeseen problems, particularly the `display` setting.  I will see if I can suggest a left intrusive set of styles.

unpacked/MathJax.js
pkra: Ah, sorry. I misunderstood the earlier discussion. Happy to do it in just mml3.js for the time being.

unpacked/extensions/MathML/mml3.js
pkra: Thanks, that would be great.

pkra: PS: Feel free to update the branch directly.

dpvc:
@pkra, if you are updating the branch, you should also change `FastHTML` to `PreviewHTML` since that is what we ended up deciding on as the final name for the old output mode.

pkra:
>  you should also change FastHTML to PreviewHTML 

Wrong issue? I think you mean https://github.com/mathjax/MathJax-dev/pull/14, correct?

dpvc:
OOPS, sorry, too many windows open at once.  Yes, that is the correct issue.

pkra:
I wrote earlier:

> I suspect the improvements from 2.5 never worked on older IE :-(

I actually confirmed that later but never reported here. So for the record, there are no regressions as far as I can tell only improvements.

PR #1262, opened 2015-09-12
A few menu display tweaks

dpvc:
This pull request includes:

* a few adjustments to the CSS to make the look for mouse (as opposed to keyboard) users more consistent with the current MathJax menu without interfering with the keyboard usage (I hope). 
    * The focus outline is not shown if the menu is opened by mouse (but focus is still set).
    * The focus outline is not shown for dialog boxes opened by mouse (but focus is still set).
    * The focus outline is not shown for close buttons when clicked by mouse
    * The focus is not shown on the math element after a menu or dialog is closed by mouse click (but is if it is closed by keyboard action).

    So the only significant difference for mouse users from the current UI is that clicking on a math element focuses it and shows the focus outline.  (Note that we probably need to implement something for `<maction>` elements so that they can get the focus and be activated by keyboard and screen reader users.  I also haven't checked href's within the math to see if they are accessible.)

* Localization of the "Close Window" string that is used on the close buttons (though I'm wondering if this would be better as "Close Dialog" or "Close Pop-up" since it is not actually the window that is being closed).

* A fix that makes sure `aria-` and `role` attributes show up as actual attributes in the DOM.

* A fix to HTML-CSS so that the focus highlighting will be full width when the math is full width (e.g., when there is a `\tag`, or with percentage-width tables, or with the `multline` environment).

* Fixes for handling of keyboard and menus in IE < 9.

* Fix so that SVG elements won't be focusable in IE.

* Fix Error jax so that [Math Processing Error] elements work with keyboard menus.

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -1048,10 +1048,11 @@ MathJax.HTML = {
           {def.style[id.replace(/-([a-z])/g,this.ucMatch)] = style[id]}}
       }
       MathJax.Hub.Insert(obj,def);
+      for (var id in def) {if (id.substr(0,5) === "aria-") obj.setAttribute(id,def[id])}
     }
     if (contents) {
       if (!(contents instanceof Array)) {contents = [contents]}
-      for (var i = 0; i < contents.length; i++) {
+      for (var i = 0, m = contents.length; i < m; i++) {
         if (contents[i] instanceof Array) {
           obj.appendChild(this.Element(contents[i][0],contents[i][1],contents[i][2]));
         } else if (type === "script") { // IE throws an error if script is added as a text node
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -2354,18 +2354,26 @@ MathJax.Hub = {
     //
     var errorSettings = this.config.errorSettings;
     var errorText = LOCALIZE(errorSettings.messageId,errorSettings.message);
-    var error = MathJax.HTML.Element("span",
-                 {className:"MathJax_Error", jaxID:"Error", isMathJax:true},errorText);
+    var error = MathJax.HTML.Element("span", {
+      className:"MathJax_Error", jaxID:"Error", isMathJax:true,
+      id: script.MathJax.error.inputID+"-Frame"
+    },errorText);
     //
     //  Attach the menu events
     //
     if (MathJax.Extension.MathEvents) {
-      error.oncontextmenu = MathJax.Extension.MathEvents.Event.Menu;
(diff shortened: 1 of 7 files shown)


dpvc:
I added a few more fixes based on the results of running the test suite.

dpvc:
Fixed Error jax so that [Math Processing Error] elements work with the keyboard menus.

zorkow:
I think WCAG requests that focus should be visible. But I really don't mind.

dpvc:
> I think WCAG requests that focus should be visible.

Note that the focus will be visible when the keyboard is used.  It is just if the menu or dialog is opened with the mouse that the focus is not visible.  This is is consistent with the [example](http://www.oaa-accessibility.org/examplep/menubar1/) linked from the [aria practices documentation](http://www.w3.org/TR/wai-aria-practices/#menu).  When the mouse is used to open and select the menu items, the focus doesn't show, but when keyboard is used, it does.  This seems consistent with normal practice.  I've never seen menus that show focus outlines for mouse actions, or buttons that focus when you click them.  Perhaps it differs by operating system, but it seems very strange to me to see the focus outlines during mouse activity.

zorkow:
* Localization of the "Close Window" string that is used on the close buttons (though I'm wondering if this would be better as "Close Dialog" or "Close Pop-up" since it is not actually the window that is being closed).

I also would prefer close dialog (or close alert). However, I am we should keep in mind that the Aria label is what the AT will read out in lieu of a button label. So what is more natural for people: window or dialog? Does your casual user even know what a dialog is?

dpvc:
> So what is more natural for people: window or dialog? Does your casual user even know what a dialog is?

I really don't know.  But I was worried that "close window" might suggest that it is the *browser* window that is being closed, not that a pseudo-window is being closed within the browser window.  A visually impaired user might not really know that a window-like item is open.  Do we need to use the aria dialog attributes and techniques (see [making dialogs modal](http://www.w3.org/TR/wai-aria-practices/#make_modal), [alert dialogs](http://www.w3.org/TR/wai-aria-practices/#alertdialog), and [modal dialogs](http://www.w3.org/TR/wai-aria-practices/#dialog_modal) for instance)?

dpvc:
OK, I've resolved the conflicts.

zorkow:
Good point. So maybe close dialog is the right way to go. Or, since we are not using a generic close button we might as well be more specific and say "close help dialog" and  "close about mathjax dialog". 

I believe we pretty much do everything re dialog/alert with the exception of using the document role for the message. And I am not sure that is really necessary given the tabbing order.

zorkow:
Should be good to merge now. It should also not clash with #1267.

PR #1263, opened 2015-09-12
Use less-aggressive CSS for ltr and inline-block, and do Edge check within the extension.

dpvc:
This makes the CSS specific to token elements (the only ones with text in them), so that things like math that uses `<mlabeledtr>` or tables with `width="100%"` won't be adversely affected by the change to `display:inline-block`.  Also, moves the Edge check into the extension, since there is more to do to set up a new browser category (I will handle that in MathJax.js).


pkra:
:+1:

dpvc:
The `detect-edge` branch of my fork of MathJax includes changes that implements `isEdge` (it is more convenient programmatically to use `isEdge` than `isMsEdge`, which would have required special processing during the version lookup).  If we think that should go into 2.6 (which I do), then we can use that here rather than the *ad-hoc* check for Edge.  If so, I'll merge `detect-edge` first, and then modify this and merge it.

pkra:
:+1: Let's do that.

PR #1264, opened 2015-09-14
Add some more ARIA

zorkow:
Adds some more aria. In particular 
* fixes the missing aria-checked elements
* adds separator role and orientation
* menu roles
* close button role and proper label attribute.
The last might clash with #1262


zorkow:
Darn... wrong base!

PR #1265, opened 2015-09-14
Add some more ARIA

zorkow:
Adds some more aria. In particular 
* fixes the missing aria-checked elements
* adds separator role and orientation
* menu roles
* close button role and proper label attribute.
The last might clash with #1262

### DIFF
--- a/unpacked/extensions/MathMenu.js
+++ b/unpacked/extensions/MathMenu.js
@@ -598,11 +598,13 @@
         {onmouseup: MENU.Mouseup,
          ondragstart: FALSE, onselectstart: FALSE, onselectend: FALSE,
          ontouchstart: MENU.Touchstart, ontouchend: MENU.Touchend,
-         className: "MathJax_MenuItem", menuItem: this},
+         className: "MathJax_MenuItem", role: this.role,
+         menuItem: this},
         def);
     },
 
     Create: function (menu) {
+      console.log('create');
       if (!this.hidden) {
         var def = this.Attributes();
         var label = this.Label(def,menu);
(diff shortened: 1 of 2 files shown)


unpacked/extensions/MathMenu.js
```diff
@@ -598,11 +598,13 @@
...
          ondragstart: FALSE, onselectstart: FALSE, onselectend: FALSE,
          ontouchstart: MENU.Touchstart, ontouchend: MENU.Touchend,
-         className: "MathJax_MenuItem", menuItem: this},
+         className: "MathJax_MenuItem", role: this.role,
+         menuItem: this},
         def);
     },
 
     Create: function (menu) {
+      console.log('create');
```
dpvc: This log message will need to be removed.

unpacked/extensions/MathMenu.js
```diff
@@ -898,19 +909,24 @@
...
           var item = child[i].menuItem;
-          if (item && item.variable === this.variable)
-            {child[i].firstChild.style.display = "none"}
+          if (item && item.variable === this.variable) {
+            child[i].firstChild.style.display = "none";
+          }
         }
         menu.firstChild.display = "";
         CONFIG.settings[this.variable] = this.value;
+        console.log('Changing');
```
dpvc: This log message will need to be removed.

dpvc:
I think you are right about the conflicts with the localization in #1262.  We should merge this one first (once the `console.log()` calls are removed), I'll fix the `more-menu` branch, and then we can merge that.

zorkow:
Oopss.... forgot to push the cleaned up version... sorry.

PR #1267, opened 2015-09-16
Fixes the menu bug introduced by giving separators a role.

zorkow:
Fixes problem that stopped keyboard navigation at separators.
Problem was introduced by giving separators the correct ARIA role.


pkra:
:+1: 

Live example https://rawgit.com/pkra/4c57afc70c4f1a4fb356/raw/c2a14411e4e0513dc0f052496c39a14e1ffdf363/mathjax-accessiblemenu.html

PR #1292, opened 2015-10-28
Update test/sample-dynamic.html

pkra:
Add reset for equations counters.

### DIFF
--- a/test/sample-dynamic.html
+++ b/test/sample-dynamic.html
@@ -54,6 +54,7 @@ input {margin-top: .7em}
     //
     window.UpdateMath = function (TeX) {
       QUEUE.Push(HIDEBOX,["Text",math,"\\displaystyle{"+TeX+"}"],SHOWBOX);
+      MathJax.Hub.Queue(["resetEquationNumbers",MathJax.InputJax.TeX]);
     }
   })();
 </script>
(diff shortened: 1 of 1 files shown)


test/sample-dynamic.html
```diff
@@ -54,6 +54,7 @@
     //
     window.UpdateMath = function (TeX) {
       QUEUE.Push(HIDEBOX,["Text",math,"\\displaystyle{"+TeX+"}"],SHOWBOX);
+      MathJax.Hub.Queue(["resetEquationNumbers",MathJax.InputJax.TeX]);
```
dpvc: You might as well use the `QUEUE` variable that is already set up rather than going to `MathJax.Hub.Queue` directly (actually, this code is from before `MathJax.Hub.Queue` existed, I think, and you had to us the `MathJax.Hub.queue` by hand).

In fact, making likes 56 and 57 be

```
QUEUE.Push(
    HIDEBOX,
    ["resetEquationNumbers",MathJax.InputJax.TeX],
    ["Text",math,"\\displaystyle{"+TeX+"}"],
    SHOWBOX
);
```

might be the best approach.

pkra:
Done.

pkra:
Thanks.

Issue #1316, opened 2015-11-27
SVG output drops a non-breaking space in mtext


pkra:
[From the User Group](https://groups.google.com/forum/#!topic/mathjax-users/Se2mxUysWtk).

The nbsp in `$\text{$x$ y}$` is lost but not in `$\text{$x$ y z}$`.  This only affects the SVG output.

dpvc:
Actually `\text{ x}` is sufficient to show the issue.

poulter7:
Hi Peter, Davide

I attempted to take a look at this issue. It seems like a problem with the old code adjusting the svg width, not having quite the desired effect here. I wasn't able to get the full test suite working from MathJax-Test, so I've not ran any regression.

pkra:
Thanks, Jonathan. Small warning: review will be a bit slow (end of year etc).

poulter7:
No problem 😊

dpvc:
@poulter7, It turns out that the real problem was severe lines lower, where nested groups are removed to simplify the SVG structure.  The spacing was being lost.  I've made a commit that fixes that problem rather than introduce the unneeded rectangle element.  See the commit listed above.

poulter7:
I'll check it out, thanks Davide

dpvc:
=> Merged.

dpvc:
=> In testsuite

MathMLToDisplay/Presentation/TokenElements/mtext/issue1316.html

PR #1318, opened 2015-11-29
Pad box in place of width adjustment in text environment

poulter7:
https://github.com/mathjax/MathJax/issues/1316 everything seems okay,
haven’t manage to get the full test suite to run yet.


pkra:
Thanks, Jonathan. Again, small warning: review will be a bit slow (end of year etc).

dpvc:
See #1316 for details, but it turns out that the real issue is in the code near the bottom of that function that removes redundant `<g>` elements when there is only a single printable character, in which case the horizontal spacing is lost.  I've made a separate commit that fixes that issue rather than introduce the unneeded rectangle.  So I'm not planning to merge this ill request, though we really do appreciate your work in trying to resolve the problem.

Issue #1314, opened 2015-11-23
HTML-CSS output cuts off very long input


pkra:
From the [User Group](https://groups.google.com/forum/#!topic/mathjax-users/1yn9UyjZZ4o).

Something very (very) long (like the result of 1000! below) has an invisible end segment. 

This does not occur on SVG or CommonHTML. 

```LaTeX
[...]

PR #1319, opened 2015-11-29
Clipping bug with large number, use auto in clip rect?

poulter7:
Proposing a fix for https://github.com/mathjax/MathJax/issues/1314

There seems no reason to clip to 1,000 in the span, maybe you could
just auto clip, which should clip to the content. Alternatively those
numbers could just be very high?

### DIFF
--- a/unpacked/jax/output/HTML-CSS/jax.js
+++ b/unpacked/jax/output/HTML-CSS/jax.js
@@ -1249,8 +1249,8 @@
         if (!this.msieClipRectBug && !bbox.noclip && !noclip) {
           var dd = 3/this.em;
           var H = (bbox.H == null ? bbox.h : bbox.H), D = (bbox.D == null ? bbox.d : bbox.D);
-          var t = HH - H - dd, b = HH + D + dd, l = 'auto', r = 'auto';
-          span.style.clip = "rect("+l+" "+r+" "+this.Em(b)+" "+this.Em(l)+")";
+          var t = HH - H - dd, b = HH + D + dd, l = -1000, r = 1000;
+          span.style.clip = "rect("+this.Em(t)+" "+this.Em(r)+" "+this.Em(b)+" "+this.Em(l)+")";
         }
       }
       // Place the box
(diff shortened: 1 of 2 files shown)


unpacked/jax/output/HTML-CSS/jax.js
```diff
@@ -1249,8 +1249,8 @@
         if (!this.msieClipRectBug && !bbox.noclip && !noclip) {
           var dd = 3/this.em;
           var H = (bbox.H == null ? bbox.h : bbox.H), D = (bbox.D == null ? bbox.d : bbox.D);
-          var t = HH - H - dd, b = HH + D + dd, l = -1000, r = 1000;
-          span.style.clip = "rect("+this.Em(t)+" "+this.Em(r)+" "+this.Em(b)+" "+this.Em(l)+")";
+          var t = HH - H - dd, b = HH + D + dd, l = 'auto', r = 'auto';
+          span.style.clip = "rect("+this.Em(t)+" "+r+" "+this.Em(b)+" "+l+")";
```
dpvc: There's really no need for the `l` and `r` variables in this case, so why not just use

    span.style.clip = "rect("+this.Em(t)+" auto "+this.Em(b)+" auto)";

directly?

dpvc:
Since this commit is based off the branch that includes the padding rectangle from your other pull request that isn't being merged, we won't be able to merge this one either.  You will need to make a new branch from the current `develop` branch and either reapply the changes by hand, or rebase the one commit, or cherry-pick it, onto the new branch.

See also the comment about `l` and `r` that I made in code diff.

dpvc:
This pull request is based off your branch that includes the mtext padding fix, so since we aren't merging those changes, this pull request also can't be merged.  You will need to make a fresh branch off of `develop` and either make the changes for the last commit again, rebase the last commit onto the new branch, or cherry-pick the commit onto the new branch.

Also,

pkra:
@poulter7 could you also sign https://www.mathjax.org/cla.html, please?

poulter7:
Thank you. I'll check this out and resubmit a pull request.

dpvc:
It turns out that using "auto" isn't going to work, as that may clip things that should be shown (e.g., the results of `\rlap{XYZ}`).  So I guess the answer is to use larger values.  I had through 1000em was large enough, but apparently not.  It's probably sufficient for the left, but not the right.  Maybe it should be `bbox.rw + 1000`?

dpvc:
I've made a commit that does as I suggested above (uses the box width plus 1000), so I'm closing this pull request.

PR #1322, opened 2015-11-30
Update AssistiveMML

### LINKED ISSUE #1235: AssitiveMML extension: visual rendering not aria-hidden in IE10+ with NVDA
Just to keep this separate from #938

pkra:
Add `role=presentation` to root element and parent of the plain MathML

Fixes #1235

### DIFF
--- a/unpacked/extensions/AssistiveMML.js
+++ b/unpacked/extensions/AssistiveMML.js
@@ -98,8 +98,9 @@
             isMathJax: true, className: "MJX_Assistive_MathML"
           });
 	  span.innerHTML = mml;
+	  frame.setAttribute("role", "presentation");
           frame.firstChild.setAttribute("aria-hidden","true");
-          span.setAttribute("aria-readonly","true");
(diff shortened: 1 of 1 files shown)


dpvc:
Are you satisfied with the tests with screen readers?  If you are ready to merge, I'll do so.

pkra:
@dpvc I think so. Do you think we can do something about the clicking issue described at  https://github.com/mathjax/MathJax/issues/1235#issuecomment-161509997 in 2.6?

dpvc:
I don't think we can center the assistive MathML, but we can probably position it a bit better.  Let me see what I can do.

dpvc:
Of course, the MathML is inside a span with clip set to a 1px by 1px square, so technically it shouldn't be clickable at all, and the measurements of the size of the MathML shouldn't matter.

pkra:
>  Let me see what I can do.

Ok. Merge this PR as you see fit then.

Issue #788, opened 2014-04-12
Implement Inline Source Renderer for Accessibility


camlorn:
I am a visually impaired computer science student close to graduation who is attempting to self-study mathematics, particularly linear algebra and differential equations.  As it currently stands, MathJax is inaccessible to me due to the basically nonexistent support for MathML.  The upshot of this is that all the pages written with MathJax are inaccessible.  I am aware of my options including mathPlayer, and if necessary I will explain exactly why it fails to work for advanced mathematics.
Regardless, MathJax can display source for me.  Unfortunately, the process to get to this source must be done per-equation and involves at least 3 advanced functions of my screen reader and at least 10 keystrokes.  I would like to be able to set MathJax to display source instead of rendered content inline so that I can actually make progress on such sites as [this linear algebra textbook](http://linear.ups.edu/html/section-SSLE.html) and Stack Exchange.  My preferred format is LaTeX, and it is my understanding that it is a somewhat common skill among blind STEM majors to be able to at least read it (Wikipedia alt text/article source for example is a great resource for us).
I believe that implementing a renderer that simply displays the source is not a major change; the time it takes to start showing up on actual web sites obviously might be.  MathML might *finally* be going somewhere, but I don't expect to see good support for it for another year at least. Advanced mathematics will be the last to be made fully functional.  This would make a very wide variety of things accessible that are not currently, opening all sorts of avenues for me and other people in my situation.  It should work on all platforms, regardless of MathML support: I don't expect linux will have it for a while, for example, and some phones probably will never will.  I would like to at least discuss doing this.  I am open to attempting to implement this myself and probably have the programming skill required, but I am clearly unable to test any visual aspects that I might or might not break in the process.
The above is clearly useful for accessibility, but may also be useful for debugging.  I am not sure on this, but it might be useful to some for that alone.
If this is the inappropriate venue for this discussion, I would appreciate being pointed in the right direction.

pkra:
This was fixed in v2.6 via the new PlainSource output, https://github.com/mathjax/MathJax/pull/1329

PR #1329, opened 2015-12-08
PlainSource output jax

pkra:
Adds a new output jax "PlainSource" which exposes a "plain text" source when possible. 

More precisely, this jax creates the following output:

* TeX or asciimath if either is used as input
* TeX or asciimath if MathML is used as input and includes annotationXML for either of these. If both are found, TeX is preferred.

This also fixes https://github.com/mathjax/MathJax/issues/788

### DIFF
--- a/unpacked/jax/output/PlainSource/jax.js
+++ b/unpacked/jax/output/PlainSource/jax.js
@@ -1,0 +1,244 @@
+/*************************************************************
+ *
+ *  MathJax/jax/output/PlainSource/jax.js
+ *
+ *  Implements the PlainSource OutputJax that displays whatever
+ *  source there was, for assistive technology users who prefer this.
+ *
+ *  ---------------------------------------------------------------------
+ *
+ *  Copyright (c) 2015 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


pkra:
@dpvc I've tried to make it IE8 compatible but I still run into issues with TeX input where I get a Math processing error (but nothing productive from lastError -- just unkown runtime error).

dpvc:
Do you want this to be in 2.6?  I don't remember if we discussed that or not.

pkra:
I wouldn't mind. Haven't got to the tests yet. sorry.

dpvc:
I have made a [pull request](https://github.com/pkra/MathJax/pull/1) to your branch to fix some issues.

unpacked/jax/output/PlainSource/jax.js
```diff
@@ -0,0 +1,236 @@
...
+    Remove: function(jax) {
+      var span = document.getElementById(jax.inputID + "-Frame");
+      if (span) {
+        if (jax.PlainSource.display) {
+          span = span.parentNode
+        }
+        span.parentNode.removeChild(span);
+      }
+      delete jax.PlainSource;
+    },
```
dpvc: BTW, the comma here is probably the thing that was tripping up IE8.  Modern browsers are OK with extra commas like this, but older IE versions choke.

pkra: Thanks for those!

pkra:
I merged your PR into my branch.

PR #1363, opened 2016-01-29
Adds configuration for excluding MathJax elements from tab order.

zorkow:
Adds a configuration parameter for including/excluding elements from the TabOrder.
The parameter is true/false, hence the actual tabIndex is returned with a function, rather than taken from a field.

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -2464,7 +2464,8 @@ MathJax.Hub = {
   },
 
   getTabOrder: function() {
-    return this.config.menuSettings.inTabOrder ? 0 : -1},
+    return this.config.menuSettings.inTabOrder ? 0 : -1;
+  },
 
   // Old browsers (e.g. Internet Explorer <= 8) do not support trim().
   SplitList: ("trim" in String.prototype ?
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -1880,7 +1880,7 @@ MathJax.Hub = {
     showMathMenu: true,      // attach math context menu to typeset math?
     showMathMenuMSIE: true,  // separtely determine if MSIE should have math menu
                              //  (since the code for that is a bit delicate)
-
+    
     menuSettings: {
       zoom: "None",        //  when to do MathZoom
       CTRL: false,         //    require CTRL for MathZoom?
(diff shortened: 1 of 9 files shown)


unpacked/MathJax.js
```diff
@@ -1896,6 +1896,7 @@ MathJax.Hub = {
       mpContext: false,    //  true means pass menu events to MathPlayer in IE
       mpMouse: false,      //  true means pass mouse events to MathPlayer in IE
       texHints: true,      //  include class names for TeXAtom elements
+      inTabOrder: false,    // set to false if math elements should be included in the tabindex
```
dpvc: Should this be `true` by default?

unpacked/extensions/MathMenu.js
```diff
@@ -1534,7 +1542,8 @@
           ITEM.RADIO("PlainSource","renderer", {action: MENU.Renderer, value:"PlainSource"}),
           ITEM.RULE(),
           ITEM.CHECKBOX("Fast Preview", "FastPreview"),
-          ITEM.CHECKBOX("Assistive MathML", "assistiveMML", {action:MENU.AssistiveMML})
+          ITEM.CHECKBOX("Assistive MathML", "assistiveMML", {action:MENU.AssistiveMML}),
+          ITEM.CHECKBOX("Include in Tab Order", "inTabOrder", {action:CONFIG.inTabOrder})
```
dpvc: Here, the `action` should be a function that gets called when the checkbox is changed.  It should add/remove the tabindex from all existing math output so that if the user changes the setting, it will take effect immediately, not just when the page is reloaded.  I would recommend that this uses `MathJax.Hub.Queue()` to queue a function that does the `Activate()` or `Unfocus()` menu actions, but without setting `MENU.posted`.

pkra:
@dpvc I spoke with @zorkow; he's ok with you going ahead and fixing it up.

zorkow:
Yes, please go ahead. I did not really get round to do much lately.

dpvc:
OOPS!  Didn't catch that this was against `master`.  I'll have to undo it.

dpvc:
I can't seem to revert the change because the master branch is protected.  I need to change that temporarily in order to revert the change, but I don't seem to be an administrator for the MathJax account, so I can't do that.  @pkra, can you make the change?

Issue #1346, opened 2015-12-31
improve npm publication


pkra:
Pushing to npm (see #1328) was botched and we un-published 2.6.0. Sigh. 

Let's improve this using CI.

wil93:
+1

For the moment I'm using this workaround

```json
"dependencies": {
  "mathjax": "github:mathjax/MathJax#master"
}
```

But this causes `npm install` to take a lot of time.

pkra:
FYI, we're wrapping up 2.6.1 https://github.com/mathjax/MathJax/milestones/MathJax%20v2.6.1

pkra:
Succesfully tested.

dpvc:
Great!

PR #1373, opened 2016-02-04
npm publish via continuous integration

pkra:
* Work towards resolving #1346
* package json: add empty test (run by travisCI by default)
* add minimal .travis.yml
  * restrict build on master branch (and PRs)
  * add npm deployment via TravisCI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: node_js
+node_js:
+  - '5'
+sudo: false
+script:
+  - npm install
+  - npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ script:
 branches:
   only:
     - master
-    - /^(\d+\.)+(\d+\.)+(\d+)?(-)?(.+)$/
+    - /^\d+\.\d+/
 deploy:
   provider: npm
   email: manager@mathjax.org
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,6 @@ sudo: false
 script:
   - npm install
   - npm test
+branches:
+  only:
+    - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,10 @@ script:
 branches:
   only:
     - master
+deploy:
+  provider: npm
+  email: manager@mathjax.org
+  api_key:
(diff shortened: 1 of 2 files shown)


pkra:
Some notes.

* Travis docs
  * https://docs.travis-ci.com/user/languages/javascript-with-nodejs
  * https://docs.travis-ci.com/user/deployment/npm
  * https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches
* Tested using https://github.com/pkra/mj-tester/, https://travis-ci.org/pkra/mj-tester/

* We should do make a pre-release on npm to test this in the real-world. 
  * I think for this particular pre-release it's enough to match versions between the tag and the package.json (e.g., 2.6.1-rc.1), not MathJax.js's internal version APIs. What do you think?

pkra:
Otherwise this particular PR is ready for review.

pkra:
I just pushed one more commit. The regex should match release tags (hopefully), e.g., 2.6.1 but also 2.6.1-rc.1, 2.7.0-beta.1 etc., but not our usual branch names. Hm... Maybe just check for starting digits?

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+node_js:
+  - '5'
+sudo: false
+script:
+  - npm install
+  - npm test
+branches:
+  only:
+    - master
+    - /^(\d+\.)+(\d+\.)+(\d+)?(-)?(.+)$/
```
dpvc: The regex is a bit strange for a couple of reasons.  In `(\d+\.)+(\d+\.)+`, only one of the two need to be repeated, since the first one will grab all the repeats and the second will get only one (since it requires at least one).  Also, the two optional ones `(\d+)?(-)?` do nothing, since the `(.+)` will match whatever them anyway.  And if you leave off the `$`, there is no need for the `(.+)`, since you are really just doing a prefix match.

Would  `/^\d+\.\d+/` be sufficient?

pkra: > Or are the capture groups used by travis in some way?

No, this was just something from my poor regex-fu (ok,fine, I admit it was google-fu). Therefore,
 
> Would /^\d+\.\d+/ be sufficient? 

sounds good. I'll update the PR later.

dpvc: It should be `/^\d+\.\d+/` (escaped `.`).  Not sure how that got lost in the copying.

pkra:
I updated the PR (hopefully without further copy&paste issues).

PR #1375, opened 2016-02-06
Revert "Adds configuration for excluding MathJax elements from tab order."

dpvc:
Reverts mathjax/MathJax#1363


dpvc:
I've used this to revert the changes for now, but I would prefer to force the master branch back to the pre-merge state.  @pkra, for that I need to either be an administrator for the MathJax account, or have you unprotect the master branch temporarily while I make the change.

pkra:
Good to go.

dpvc:
OK, it is put back the way it was.  Thanks!

Issue #1352, opened 2016-01-20
Make `tabindex=0` optional


pkra:
As per F2F.

We will add a menu option to switch off `tabindex=0` on MathJax output. This will allow users to disable it should they want to.

The Menu structure will be: Math Settings => Renderer => checkbox "Include in tab order" (below the AssistiveMML checkbox)

The docs will strongly discourage developers from initially disabling this (by means of configuration) as it will render the menu inaccessible.

pkra:
From https://github.com/mathjax/MathJax/issues/893#issuecomment-173273238

> it's not that I want to remove it necessarily, it's more to be able to manage it myself. If I have a list with tabindex 1,2,3,4 and MathJax comes in the middle and places a 0 giving 1,0,2,3,4 - it is very difficult to manage this

We should consider this use case.

pkra:
> We should consider this use case.

We will not provide any internal functionality for this. Instead, we suggest to hook into the `new math` signal since it passes the actual element name and thus allows developer to then set the tabindex as they see fit.

pkra:
See also https://github.com/mathjax/MathJax/pull/1363

dpvc:
This is handled by PR 1376.

dpvc:
=> In test suite.

Configuration/MathMenu/tabindex-1.html
Configuration/MathMenu/tabindex-2.html
Configuration/MathMenu/tabindex-3.html

PR #1376, opened 2016-02-06
Tabindex

dpvc:
Do Volker's tabindex changes, but from and to the `develop` branch rather than `master`.  :-)

I changed the default to `inTabOrder: true`, and refactored the code that sets all the `tabIndex` values so that they can be called by the page author to remove/set all the indices if they want.  I removed the menu item `action` value, since selecting any menu item causes the tabindices to be reset, so no special action needs to be take for this menu item.  I left the `getTabOrder()` function, since it includes logic that is needed to get the index from the menu setting, and using the function prevents that from being duplicated everywhere.  As a side-effect, a page author could override it, but that is not its main purpose.  I added a `jax` parameter to `getTabIndex()` so that, if overridden, it could be used to identify which math element was being processed.

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -2368,14 +2368,14 @@ MathJax.Hub = {
       error.oncontextmenu = EVENT.Menu;
       error.onmousedown = EVENT.Mousedown;
       error.onkeydown = EVENT.Keydown;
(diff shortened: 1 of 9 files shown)


dpvc:
Resolves issue #1352

PR #1394, opened 2016-02-19
feat(color-extension): add class names

barillax:
We have a custom extension that sometimes sets the color of containing elements to green (in addition to some other interactive behavior it adds). However, we wanted to be able to defer to overridden colors that were set by the `color.js` extension. In pseudocode:

`$ x + \mycustomextension[y]  $`   

> `y` is green as a side effect of \mycustomextension

`$ x + \color{red} \mycustomextension[y] $`

> `y` is red due to \color

To accomplish this, we needed `color.js` to set a class, `mjx-extension-color` on the `.mjx-mstyle` element that it creates so that we could adjust our CSS to `color: inherit` when `.mjx-extension-color` is an ancestor. Without this class, we can't distinguish between overridden colors and overridden sizes, for example.

This PR includes the simple changes to `color.js` to accommodate this. I'm open to suggestions for a better class name than `mjx-extension-color`.

### DIFF
--- a/unpacked/extensions/TeX/color.js
+++ b/unpacked/extensions/TeX/color.js
@@ -223,7 +223,7 @@ MathJax.Hub.Register.StartupHook("TeX Jax Ready",function () {
       var model = this.GetBrackets(name),
           color = this.GetArgument(name);
       color = COLOR.getColor(model,color);
-      var mml = STACKITEM.style().With({styles:{mathcolor:color}});
+      var mml = STACKITEM.style().With({styles:{mathcolor:color, "class": "mjx-extension-color"}});
       this.stack.env.color = color;
       this.Push(mml);
     },
@@ -235,7 +235,7 @@ MathJax.Hub.Register.StartupHook("TeX Jax Ready",function () {
       var old = this.stack.env.color; this.stack.env.color = color;
       var math = this.ParseArg(name);
       if (old) {this.stack.env.color} else {delete this.stack.env.color}
-      this.Push(MML.mstyle(math).With({mathcolor: color}));
+      this.Push(MML.mstyle(math).With({mathcolor: color, "class": "mjx-extension-color"}));
     },
 
     //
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
Just a note that it might be a good pattern to have a best practices to have all plugins provide some identifier (class or attribute) on elements which get modified as a result of the plugin processing.

pkra:
Please make a new PR based off `develop`. See https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md#working-on-mathjax-core. Also, we'd need a signed copy of https://www.mathjax.org/cla/ but you might want to wait until this change is reviewed.

dpvc:
Here is an easy change you can add as part of your extension that will accomplish the same thing without needing to change MathJax's core.

```
MathJax.Hub.Register.StartupHook("TeX color Ready",function () {
  var MML = MathJax.ElementJax.mml;
  var STACKITEM = MathJax.InputJax.TeX.Stack.Item;
  var PUSH = STACKITEM.prototype.Push;
  
  STACKITEM.Augment({
    Push: function (mml) {
      if (mml instanceof MML.mbase && (mml.mathcolor || mml.mathbackground)) {
        if (mml["class"]) {
          if (!mml["class"].match(/mjx-extension-color/)) mml["class"] += " mjx-extension-color";
        } else mml["class"]  = "mjx-extension-color";
      }
      return PUSH.apply(this,arguments);
    }
  });
});
```

This will mark any TeX input that produces MML with `mathcolor` or `mathbackground` attributes with class `mjx-extension-color`.  If you want to exclude colors that you set yourself, you could add more logic to eliminate those (e.g., via a flag you set on the elements you create; without more details about your extension, I can't tell what might be needed).

Because this is a very special case need, I'm against merging the pull request, especially if the code above can satisfy your needs.

pkra:
@dpvc wrote 
> Because this is a very special use-case need, I'm against merging the pull request, especially if the code above can satisfy your needs.

:+1:

@booleanbetrayal wrote

> Just a note that it might be a good pattern to have a best practices to have all plugins provide some identifier (class or attribute) on elements which get modified as a result of the plugin processing.

While I like the idea of having more information in the output, I suspect this would quickly get out of hand (and is not always possible from a TeX input).

booleanbetrayal:
Totally make sense @pkra! The `PUSH` modification works like a charm, @dpvc. Thanks so much!

PR #1395, opened 2016-02-19
Don't die if script has been removed from document

uhoreg:
On our site, which displays a live preview of math as users type it in, it appears that if users type too fast, MathJax will die and not display any more math.  In the console, it complains that "b.parentNode is null".  This patch checks b.parentNode (a.k.a. script.parentNode in the non-minified version) before calling b.parentNode.insertBefore().

I suppose it would be better to check script.parentNode earlier in the process so that MathJax doesn't waste time processing something that won't be displayed, but this patch seems to work, and should probably be done anyways.

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -2380,7 +2380,7 @@ MathJax.Hub = {
     //
     var node = document.getElementById(error.id);
     if (node) node.parentNode.removeChild(node);
-    script.parentNode.insertBefore(error,script);
+    if (script.parentNode) script.parentNode.insertBefore(error,script);
     if (script.MathJax.preview) {script.MathJax.preview.innerHTML = ""}
     //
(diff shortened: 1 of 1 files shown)


dpvc:
I'm OK with the requested change, but the pull request should be for a branch that branched off of our `develop` branch (not the `master` branch) and that is directed to the `develop` branch (not the `master` branch).  See the documentation on [contributing](https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md#key-branches--tags) to mathjax for details.  

You could rebase your patch to make it branch off `develop`, and then create a new pull request.  Or just make a new branch by hand, since the change is easy.

In the meantime, I'm closing this request, since it from and to the wrong branch.

pkra:
We'd also need https://www.mathjax.org/cla/ signed. Ping me by email if you have any questions.

PR #1448, opened 2016-04-21
[CommonHTML] fix typo in styles

### LINKED ISSUE #1443: echidna CSS validation issues
Reported at https://github.com/w3c/echidna/issues/271

Bug: CHTML has a typo `".mjx-span":   {display:"span"}`  https://github.com/mathjax/MathJax/blob/d4ab1b35c96dd964eaa9e1ed2c86e39fffbdacf6/unpacked/jax/output/CommonHTML/jax.js#L126.

Not strictly a bug: but the menu still uses `DXImageTransform` etc (which are strictly speaking invalid). Not sure if we want to drop them but I feel like we could.

pkra:
Fixes #1443

### DIFF
--- a/unpacked/jax/output/CommonHTML/jax.js
+++ b/unpacked/jax/output/CommonHTML/jax.js
@@ -123,7 +123,7 @@
 
     ".mjx-box":    {display:"inline-block"},
     ".mjx-block":  {display:"block"},
-    ".mjx-span":   {display:"span"},
+    ".mjx-span":   {display:"inline"},
     ".mjx-char":   {display:"block", "white-space":"pre"},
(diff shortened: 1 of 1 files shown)


pkra:
On the other hand, this style is redundant, no?

dpvc:
> On the other hand, this style is redundant, no?

No.  A few lines earlier, you see

```
".mjx-math *": {display:"inline-block", "text-align":"left"},
```

which makes everything inside the CHTML output be `inline-block` by default.  So this `mjx-span` class allows you to get back to regular `inline` elements when needed.  But apparently that never actually worked.

pkra:
> No. A few lines earlier, you see

Thanks, I had missed that!

PR #1453, opened 2016-05-03
CommonHTML: Pass full font URL to Ajax.fileURL

kovidgoyal:
Rather than just passing the base font dir, pass the
full font file path. This allows for user code that overrides
fileURL() to properly map font file URLs as well.

I need this because in my application I am loading MathJax from IndexedDB, which means all mathjax files become blob urls, and my application has to map all MathJax URLs to Blob URLs, which it does by monkeypatching Ajax.fileURL() in the AuthorInit() function. See https://github.com/kovidgoyal/calibre/blob/master/src/pyj/read_book/mathjax.pyj#L15

### DIFF
--- a/unpacked/jax/output/CommonHTML/fonts/TeX/fontdata.js
+++ b/unpacked/jax/output/CommonHTML/fonts/TeX/fontdata.js
@@ -1641,9 +1641,6 @@
   };
   (function () {
     var STYLES = CHTML.config.styles, FONTS = CHTML.FONTDATA.FONTS;
-    var OTFDIR = AJAX.fileURL(CHTML.webfontDir+"/TeX/otf"),
-        EOTDIR = AJAX.fileURL(CHTML.webfontDir+"/TeX/eot"),
-        WOFFDIR = AJAX.fileURL(CHTML.webfontDir+"/TeX/woff");
     var faces = [];
     for (var name in FONTS) {if (FONTS.hasOwnProperty(name)) {
       var family = CHTML.FONTDATA.familyName(name), FAMILY = family;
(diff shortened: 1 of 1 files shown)


pkra:
Closed only for technical reasons -- PRs must go to to the `develop` branch. See also https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md#submitting-pull-requests

pkra:
In particular, we'd need https://www.mathjax.org/cla/ but maybe wait until @dpvc comments on the actual code :smile:

kovidgoyal:
I am not particularly attached to the code. All I care about is having some way to implement my end goal -- which is user controlled mapping of web font URLs. So if you agree with this patch let me know and I'll jump through whatever hoops you need for accepting PRs. Otherwise, I'll open a bug report asking for this feature and let you implement it however you like.

pkra:
Thanks so much, @kovidgoyal! Let's see what Davide writes.

dpvc:
I guess I'm OK with the changes in principle, but I'd want to do it very slightly differently, and also do the corresponding thing in the HTML-CSS `@font-face` creation as well (for consistency).  I take it you are dropping the `.eot` and `.otf` references, which is why you are using `if (src) ...`.  The current `fileURL()` method never returns falsey values, so that test is redundant as far as MathJax is concerned.  I feel a little funny adding code just for one person's application, but I guess it doesn't hurt too much.

Since I want to add the same to HTML-CSS, why don't I make the changes and we won't have to bother with a CLA from you.

kovidgoyal:
Fine by me. Yes, I am dropping those, as otherwise it leads to error messages in the console when the security policy in the browser prevents those files from being loaded 9I have MathJax running in an isolated iframe).  If you dont like it, I suggest making it a separate option in the CommonHTML backend that defaults to [eot, woff, otf]. 

Do you want me to open a bug report?

pkra:
> I feel a little funny adding code just for one person's application, but I guess it doesn't hurt too much.

Let's discuss this at the next F2F?

eli-schwartz:
What ever happened to this? @kovidgoyal's fork is still used for calibre, and is now 477 commits behind this repository, would be nice if this was upstreamed and then general improvements automatically made use of by calibre.

dpvc:
Nothing was done about it.  There is, however, a way to handle this without the need for changes to MathJax itself.  You can use

```
(function () {
  var StyleString = MathJax.Ajax.StyleString.bind(MathJax.Ajax);
  MathJax.Ajax.StyleString = function (styles) {
    return StyleString(styles).replace(/url\('?(.*?)'?\)/g, function (match, url) {
      //  ... translage url to whatever you want ...
      return url;
    });
  };
})()
</script>
```

to trap and replace all the `url()` references in the stylesheets.  Place this just *before* the script that loads `MathJax.js` itself.

kovidgoyal:
@dpvc that is not sufficient. calibre runs mathjax inside a sandboxed iframe, which means all URLs, including those for other js files have to be re-written so that they can be loaded from within the sandbox, because of CSP restrictions.

dpvc:
I had expected that you would continue to have a modified `MathJax.Ajax.fileURL()` to handle those other cases.  Your pull request only deals with the font URLs, so that is what I provided above.  If you did

```
(function () {
  var StyleString = MathJax.Ajax.StyleString.bind(MathJax.Ajax);
  MathJax.Ajax.StyleString = function (styles) {
    return StyleString(styles).replace(/url\('?(.*?)'?\)/g, function (match, url) {
      return MathJax.Ajax.fileURL(url);
    });
  };
})()
</script>
```

I think that would be equivalent to your pull request.  Of course, you would still want to override the `fileURL()` method in order to get whatever URL translation you need.  You can do that in a similar way to the above, so you should be able to get the adjustments you need through configuration, without having to modify MathJax itself.

kovidgoyal:
ah ok that makes sense, thanks :)

Issue #831, opened 2014-05-30
Invisible font in SVG Rendering


Herst:
Odd issue, reproduced on latest Firefox on Windows (/edit: and Linux) with MathJax 2.3 and 2.4 on one website (but not e.g. the MathJax Play Area) using my bookmarklet:

Steps to reproduce:
1.) Go to http://www.informatik-forum.at/showthread.php?t=107276 (inofficial forum of my university, thread where I present my bookmarklet that brings MathJax power to there for those interested)
2.) In the first box after "Code:" you will see a bookmarklet, press F12 to go into the JS console and execute it there (interestingly no need to remove the "javascript:" prefix) (alternatively save it as a bookmarklet and then run it or use the unminimized code from the second box).
3.) The MimeTex-generated images will be replaced by MathJax-generated content. Change the rendering mode to SVG.
4.) Then it looks the following (at least for me):
![mj_svg_bug](https://cloud.githubusercontent.com/assets/1042176/3135541/bb6012ca-e834-11e3-8e81-72599e003b46.png)
(for CDN version now, for beta replace the domains in the bookmarklet code)

Like I said reproduced on latest Firefox on Windows 8.1 (and latest Ubuntu). Fine on latest Google Chrome and and IE11 on Win8.1 (although in former I get a probably unrelated bug with the zoom window albeit also just in SVG mode: ![mj_svg_zoom_bug](https://cloud.githubusercontent.com/assets/1042176/3135579/3789b554-e835-11e3-8a35-a423bd787016.png) [should I fill extra bug report?])

PS: Notice also the lonely Eszett ("ß") that plays a role in bug #826 but the bug doesn't show here because STIX has that character (I guess).

PPS: Feedback on my bookmarklet is always appreciated.

pkra:
I cannot reproduce the OP with 2.5.1. @Herst can you?

I'll create a separate bug for the about box issue.

Herst:
@pkra Yes I can. Same procedure as described in the first post: ran Bookmarklet from there, changed to SVG and voila! (Firefox 37b2, with some add-ons and not a clean profile though)

pkra:
Strange. Now I see the problem again. I could swear it was working 2 days ago... bummer.

episodeyang:
I'm running into the same problem.

Herst:
@episodeyang With a different testcase? If yes then please share it with us.

pkra:
[...]

PR #1454, opened 2016-05-03
Use full location URL for xlink references in SVG <use> elements

dpvc:
Use full location URL for xlink references in SVG `<use>` elements when the page includes a `<base>` element.  Resolves issue #831.

The might cause issues when the page URL uses query parameter (i.e., ends with `?...`).  For situations where that will occur, the use of `useFontCase: false` is recommended.

### DIFF
--- a/unpacked/jax/output/SVG/jax.js
+++ b/unpacked/jax/output/SVG/jax.js
@@ -37,8 +37,8 @@
   //  Get the URL of the page (for use with xlink:href) when there
   //  is a <base> element on the page.
   //  
-  var SVGURL = String(document.location).replace(/#.*$/,"");
-  if (document.getElementsByTagName("base").length === 0) SVGURL = "";
+  var SVGURL = (document.getElementsByTagName("base").length === 0) ? "" :
+                String(document.location).replace(/#.*$/,"");
 
   SVG.Augment({
     HFUZZ: 2,     // adjustments for height and depth of final svg element
--- a/unpacked/jax/output/SVG/jax.js
+++ b/unpacked/jax/output/SVG/jax.js
@@ -33,6 +33,13 @@
   var SVGNS   = "http://www.w3.org/2000/svg";
   var XLINKNS = "http://www.w3.org/1999/xlink";
 
+  //
+  //  Get the URL of the page (for use with xlink:href) when there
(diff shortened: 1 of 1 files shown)


unpacked/jax/output/SVG/jax.js
```diff
@@ -33,6 +33,13 @@
   var SVGNS   = "http://www.w3.org/2000/svg";
   var XLINKNS = "http://www.w3.org/1999/xlink";
 
+  //
+  //  Get the URL of the page (for use with xlink:href) when there
+  //  is a <base> element on the page.
+  //
```
zorkow: "when there is >no< <base> element on the page"?

unpacked/jax/output/SVG/jax.js
```diff
@@ -33,6 +33,13 @@
   var SVGNS   = "http://www.w3.org/2000/svg";
   var XLINKNS = "http://www.w3.org/1999/xlink";
 
+  //
+  //  Get the URL of the page (for use with xlink:href) when there
+  //  is a <base> element on the page.
+  //  
+  var SVGURL = String(document.location).replace(/#.*$/,"");
+  if (document.getElementsByTagName("base").length === 0) SVGURL = "";
+
```
zorkow: var SVGURL = (document.getElementsByTagName("base").length === 0) ? "" : String(document.location).replace(/#.*$/,"");

will make it more efficient.

unpacked/jax/output/SVG/jax.js
dpvc: No, it is when there is a `<base>` element, as the comment says now.  The issue is that the `<base>` element makes relative xlink references point to the wrong place (the base), so we need to use absolute references in that case (the case where there is a `<base>` element).

unpacked/jax/output/SVG/jax.js
dpvc: OK.

unpacked/jax/output/SVG/jax.js
zorkow: Right, got it.

dpvc:
I've made the suggested change.

zorkow:
lgtm.

Issue #1384, opened 2016-02-11
various issues on puzzling.SE


pkra:
From [puzzling.SE](http://meta.puzzling.stackexchange.com/questions/4551/mathjax-glitch-when-viewing-comment?noredirect=1#comment9640_4551).

On [this question](http://puzzling.stackexchange.com/questions/26226/concentrating-tokens-on-an-infinite-board/26234?noredirect=1#comment80729_26234), I see the following

* the FastPreview crashes half-way through the page (TeX is still in place). The console reports `MathJax.js?config=TeX-AMS_HTML-full:19 Uncaught TypeError: Cannot read property 'insertBefore' of null`

* If I deactivate the Fast Preview, then I see the problem described in the post -- furthe down in the page, the rendering is all over the place and equations are duplicated. This seems to roughly match the point where the FastPreview crashes (the last answer from BigBlackBox)

I'll see if later versions of Chrome suffer from this.

pkra:
CommonHTML and SVG don't seem to be affected.

pkra:
Also, I used the standard modern.ie VM for Win7, IE9 for this.

pkra:
Same on Chrome 49.0.2623.47 beta-m, Win7.

dpvc:
I can reproduce on Chrome 48.0.2564.97 on Mac OS.  It is not dependent on the Fast Preview or Assistive MathML settings for me, and I see no errors in the error log.

dpvc:
The state of the DOM is very similar to that in issue #1171.  Im wondering if there isn't a second `MathJax.Hub.Typeset` being made somewhere that doesn't use the MathJax queue.  There are several occurrences of raw `MathJax.Hub.Typeset` calls in the javascript loaded for the page, but since the code is minified, it is hard to tell when they get used.  Rerendering the page (not reloading but calling `MathJax.Hub.Rerender()` in the console) clears the problem, so it is not directly an output error.  It really looks like to Typeset calls being interleaved.

pkra:
Thanks. I'll follow up with SE.

m0sa:
Users are reporting getting the same error [here](http://meta.dsp.stackexchange.com/questions/1340/mathjax-not-rendering), I couldn't repro it with the provided steps though.

dpvc:
[...]

PR #1456, opened 2016-05-03
Fix problems from puzzling.SE

dpvc:
This fixes some places where the removal of DOM elements while MathJax is running could cause MathJax to stop processing, or to leave duplicate copies of the typeset math in place.  Resolves issue #1384.


zorkow:
Should be a lot easier with classlist in the future.
O/w lgtm.

dpvc:
> Should be a lot easier with classlist in the future.

Agreed.

Issue #1451, opened 2016-04-28
MathJax <mi mathvariant=“double-struck”> Common HTML not displaying properly in iOS and Android mobile


WebStew:
**Issue Summary**

Converting our current MathJax rendering option from HTML-CSS to commonHTML is no longer rendering mathvariant="double-struck" characters in some instances.

**Steps to Reproduce**

Visit [CodePen](http://codepen.io/Terminalpunk/pen/pyxPWV) with iPhone 6 Chrome or Samsung Galaxy S5 Chome

**Technical Details**

- MathJax Version: v2.6
- Output : CommonHTML
- Client : iPhone 6 Chrome Version 50.0 iOS 9.3.1 
- Client : Samsung Galaxy S5 Android Version 5 Chome Version 39.0

![double-struck](https://cloud.githubusercontent.com/assets/1636565/14888650/46085c94-0d53-11e6-878f-158662e7372b.PNG)

dpvc:
When you make a reference to a double-strick character like

```
<mi mathvariant="double-struck">&#x1D53D;</mi>
```

you are actually over-specifying the variant.  The character `&#x1F53D;` is already double-struck, so there should be no need to specify `mathvariant="double-struck"` in this case.  The CommonHTML should still handle this situation correctly, but currently doesn't. What's happening is that MathJax doesn't actually find this character in its fonts and so makes a Unicode reference hoping that the character can be found in some font installed on your system.  When you view this from a desktop machine, you apparently have fonts that do include these character positions (perhaps the STIX fonts?), and so the system can display them.  Most mobile devices don't have fonts for these, so they are not being displayed there.

The example that *does* work on the mobiles uses the double struck P, which is at `&#x2119;`, and it turns out that CommonHTML does properly recognize the characters in the Letterlike Symbols block where this one is found, so it is handled properly for you (but note the missing double-struck B and G in the equations including the P's).  The reason that the double-struck B (and G) works in the initial equation is that you used

```
<mi mathvariant="double-struck">B</mi>
```

rather than using the Math Alphabet reference at `&#x1D539;`.

The CommonHTML output *should* translate the Math Alphabetic references to the proper locations in its own font, but it isn't currently doing so (the HTML-CSS and SVG output do, which is why it worked for you before).  This should be fixed, but for now, you could use

```
<mi mathvariant="double-struck">F</mi>
```

rather than

```
<mi mathvariant="double-struck">&#x1D53D;</mi>
```

until we produce a fix for this.

[...]

PR #1457, opened 2016-05-03
Fix problem with Plane1 not being mapped to the MathJax fonts properly

dpvc:
This makes sure the Math Alphabetical characters are mapped down to the ASCII range so that they display properly in the MathJax fonts (which are only Plane0).  Resolves issue #1451.

### DIFF
--- a/unpacked/jax/output/CommonHTML/jax.js
+++ b/unpacked/jax/output/CommonHTML/jax.js
@@ -744,6 +744,10 @@
     getCharList: function (variant,n) {
       var id, M, list = [], cache = variant.cache, nn = n;
       if (cache[n]) return cache[n];
+      if (n > 0xFFFF && this.FONTDATA.RemapPlane1) {
+        var nv = this.FONTDATA.RemapPlane1(n,variant);
+        n = nv.n; variant = nv.variant;
+      }
       var RANGES = this.FONTDATA.RANGES, VARIANT = this.FONTDATA.VARIANT;
       if (n >= RANGES[0].low && n <= RANGES[RANGES.length-1].high) {
         for (id = 0, M = RANGES.length; id < M; id++) {


unpacked/jax/output/CommonHTML/jax.js
```diff
@@ -744,6 +744,10 @@
     getCharList: function (variant,n) {
       var id, M, list = [], cache = variant.cache, nn = n;
       if (cache[n]) return cache[n];
+      if (n > 0xFFFF && this.FONTDATA.RemapPlane1) {
+        var nv = this.FONTDATA.RemapPlane1(n,variant);
+        n = nv.n; variant = nv.variant;
+      }
```
zorkow: I still don't like double parked assignments.

zorkow:
LGTM (I don't quite understand the context, though.)

unpacked/jax/output/CommonHTML/jax.js
dpvc: Of course, this would be

    let [n,variant] = this.FONTDATA.RemapPlane1(n,variant);

Putting the two assignments on the same line is my way of imitating this.  The purpose of those two lines is to decompose the object obtained from the previous command.  It has always bothered me that JavaScript didn't have a way to do that directly, so I'm glad to see it in ES6.

dpvc:
The context is that the MathJax TeX fonts only have glyphs in Unicode plane 0.  At the time they were created, there was not very good support for Plane1, and so we could not rely on it.  Also, because we were sending these over the network, we wanted to keep them as small as possible, and so didn't want to have duplicate characters in Plane1 when we would have to use Plane0 for some browsers anyway.  To further help make the fonts small, we broke up the characters into a number of separate fonts (26 as I recall) so that only small sets of characters needed to be downloaded rather than the whole set.  The STIX fonts, for example, include nearly 9000 glyphs, and are prohibitively large to download as a web font.

The Math Alphabet characters (like the double-struck characters, and the Fraktur set) are in Plane1; e.g., double-struck are (mostly) at U-1D538 to U-1D56B.  But MathJax's fonts handle them by putting them at the usual ASCII positions, but in a separate font for the double-struck versions.  The same for Fraktur, script, sans-serif, etc.

In MathML these characters can be specified in two different ways:  Either directly as their Plane1 versions, e.g., `<mi>&#x1D538;</mi>` or via the `mathvariant`, e.g., `<mi mathvariant="double-struck">A</mi>`. Having the double-struck characters in the ASCII positions makes it easy to handle the second of these (by having a style that selects the font with the double-struck characters).  Indeed, the Firefox MathML did not support mathvariants other than normal, italic, bold, and bold-italic (I'm not sure if it does even yet), so MathJax added styles to map double-struck, Fraktur, etc. to its own fonts when NativeMML output was used.

For the direct Unicode values, since the MathJax fonts don't include the glyphs at the Plane1 locations, MathJax must convert them to their ASCII positions and apply the correct font for the alphabet in use.  That is what the code above does.  The original issue had someone using `<mi mathvariant="double-struck">&#x1D538;</mi>`, which is redundant, but MathJax *should* handle it anyway.  It does in HTML-CSS and SVG output, but didn't in CommonHTML.  The `RemapPlane1()` code was included, but I forgot to call it.  The patch above adds the missing call.

Issue #1422, opened 2016-03-24
x-height calculation incorrect in calibre viewer on windows with 2.6.0


kovidgoyal:
I recently updated the version of mathjax included with calibre to 2.6.0. However, on windows it calculates the font x-height incorrectly. linux and OS X are fine. 

I tried updating to 2.6.1 -- no change.

Note that calibre uses Qt WebKit to render HTML and if you want to play with the mathjax bundled with calibre you can find it in resources/viewer/mathjax in the calibre installation directory. It is unbundled, so it should be easy to modify and edit. A sample epub file which shows this problem is https://manual.calibre-ebook.com/_static/mathjax.epub

Simply open it in the calibre viewer and you will see that math is getting typeset at approx 50% of correct height. You can right click in the viewer and choose Inspect to get access to a javascript console. 

I did a bit of debugging and I one difference I found is that the height of the ExSpan ins ~550px on linux and 300px on windows. However, changing .MathJax_SVG_ExBox to have height 120ex instead of 60ex does not affect rendering, so I'm not sure how to proceed. If you can tell me what I need to change to affect the x height calculation, I can probably special case it for calibre on windows.

kovidgoyal:
Oh and I forgot to mention, calibre uses SVG output, so the ExSpan I am talking about is in SVG/jax.js

pkra:
What Calibre version would this be happening on? Also, what QTWebKit version are you using?

FWIW, QTWebKit has been problematic in the past since it's a rather stripped down version of WebKit so it might be hard to fix (unless you're already on the Chromium-based version, I suppose; that might have a new set of problems we haven't seen yet).

pkra:
One thing worth checking might be http://docs.mathjax.org/en/latest/output.html#viewport-meta-tag, depending on whether Calibre does that or not (the epub file does not have any meta tags for viewport size).

kovidgoyal:
Qt 5.5.1 and Qt WebKit (the old chromium engine), not Qt WebEngine. (Although that will likely change in the future, when I migrate the viewer to Qt WebEngine). 

I dont really need you to fix anything, this is almost certainly a bug in Qt WebKit (the ex unit is basically much smaller in windows than on other platforms. If you can tell me where MathJax does its x height calculations for SVG output, I will simply special case them in the version bundled with calibre to workaround this bug in Qt WebKit.

kovidgoyal:
calibre version 2.53 (the latest)

dpvc:
[...]

PR #1458, opened 2016-05-03
Fix problem with container width in SVG when math is scaled.

dpvc:
If the scaling factor for all math (in the math menu or via configuration) is not 100%, the SVG container width would be incorrectly scaled as well.  This sets the size correctly so that the scaled SVG element will be the correct width when there are equation numbers, or for line breaking.  Resolves issue #1422.


zorkow:
lgtm.

Issue #1447, opened 2016-04-20
"Math Processing Error" using CommonHTML with line breaks


hubgit:
An error is thrown when rendering a formula with the CommonHTML renderer in a column that's narrow enough to trigger line breaking, when the `linebreaks` config is set to `automatic: true`.

```
> MathJax.Hub.lastError
> TypeError: Cannot read property 'CHTML' of undefined
    at MML.msubsup.Augment.CHTMLmoveLine (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:521:37)
    at MML.mbase.Augment.CHTMLmoveSlice (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:306:12)
    at MML.mbase.Augment.CHTMLmoveLine (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:267:22)
    at MML.mbase.Augment.CHTMLmoveSlice (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:306:12)
    at MML.mbase.Augment.CHTMLmoveLine (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:267:22)
    at MML.mbase.Augment.CHTMLmoveSlice (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:306:12)
    at MML.mbase.Augment.CHTMLmoveLine (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:267:22)
    at MML.mbase.Augment.CHTMLaddLine (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:195:12)
    at MML.mbase.Augment.CHTMLmultiline (https://cdn.mathjax.org/mathjax/2.6-latest/unpacked/jax/output/CommonHTML/autoload/multiline.js?rev=2.6.1:121:12)
[...]

PR #1459, opened 2016-05-03
Improve getNode() so that is finds the correct node. #1447

dpvc:
Improves `getNode()` so that it doesn't find elements inside nodes that come from other MathML nodes (those with IDs).  Resolves issue #1447.  See that issue for examples that cause the problem.

### DIFF
--- a/unpacked/jax/output/CommonHTML/jax.js
+++ b/unpacked/jax/output/CommonHTML/jax.js
@@ -337,19 +337,28 @@
     setScript: HTML.setScript,
     
     //
-    //  Look through the children of a node for one with the given type
-    //  but don't step into child nodes that are from MathML elements
-    //  themselves (they will have IDs).
+    //  Look through the direct children of a node for one with the given
+    //  type (but if the node has intervening containers for its children,
+    //  step into them; note that elements corresponding to MathML nodes
+    //  will have id's so we don't step into them).
+    //  
+    //  This is used by munderover and msubsup to locate their child elements
+    //  when they are part of an embellished operator that is being stretched.
+    //  We don't use querySelector because we want to find only the direct child
+    //  nodes, not nodes that might be nested deeper in the tree (see issue #1447).
     //
     getNode: function (node,type) {
-      var name = RegExp("\\b"+type+"\\b");
-      for (var i = 0, m = node.childNodes.length; i < m; i++) {
-        var child = node.childNodes[i];
-        if (name.test(child.className)) return child;
-        if (child.id == null) return this.getNode(child,type);
+      while (node && node.childNodes.length === 1 && node.firstChild.id == null)
+        node = node.firstChild;
+      if (node) {
+        var name = RegExp("\\b"+type+"\\b");
+        for (var i = 0, m = node.childNodes.length; i < m; i++) {
+          var child = node.childNodes[i];
+          if (name.test(child.className)) return child;
+        }
       }
+      return null;
     },
-
 
     /********************************************/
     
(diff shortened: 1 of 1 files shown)


unpacked/jax/output/CommonHTML/jax.js
```diff
@@ -336,34 +336,20 @@
...
-      if (nodes.length === 1) return nodes[0];
-      var closest = nodes[0], N = this.getNodeDepth(node,closest);
-      for (var i = 1, m = nodes.length; i < m; i++) {
-        var n = this.getNodeDepth(node,nodes[i]);
-        if (n < N) {closest = nodes[i]; N = n}
+      var name = RegExp("\\b"+type+"\\b");
+      for (var i = 0, m = node.childNodes.length; i < m; i++) {
+        var child = node.childNodes[i];
+        if (name.test(child.className)) return child;
+        if (child.id == null) return this.getNode(child,type);
```
zorkow: Much easier to understand. No idea what the original was doing... But what if there are multiple children without id?
Another  problem I can see is that you now accumulate stack space. It's not tail recursive and in JS < ES6 it would not help anyway. Also would a query selector not be a better alternative (assuming there is no IE8 problem)?

One way to make it a bit more lightweight is to bind the name test outside the recursive context.

dpvc: All good observations, especially for a general search command.

This is not a general search command, however.  It is used (in several places) for one specific purpose:  when an embellished operator needs to have its operator to be stretched, the embellishments may need to be adjusted. For example, an `munderover` with a base that is stretched may need to have its over accent stretched as well.  In these cases, the parent element (`munderover` in this example) needs to locate the elements that it has used for the base and over nodes.  This routine allows the parent to locate those elements.  These elements have class names that are things like `mjx-base` and `mjx-over`, and those elements are containers in which the actual child MathML nodes are rendered.  Usually, they are the direct children of the element used for the `munderover`, but in some cases, there is an intervening element.

Al the nodes that correspond to the original MathML elements are given ids, but the `mjx-base` and other such sub-parts of the `munderover` do not.  So the test looks through the direct children of the given element for one with the correct type, but if the child has no id (there will be only one in that case), we look for *its* children instead.

In practice, the routine will only recurse once (at most), so I'm not too worked about tail recursion or the extra `RegExp` call here.  But I suppose it could be done more directly as

```
getNode: function (node,type) {
  while (node && node.childNodes.length === 1 && node.firstChild.id == null) node = node.firstChild;
  if (node) {
    var name = RegExp("\\b"+type+"\\b");
    for (var i = 0, m = node.childNodes.length; i < m; i++) {
      var child = node.childNodes[i];
      if (name.test(child.className)) return child;
    }
  }
  return null;
}
```

which avoids the recursion all together.

dpvc: PS, I originally did use a query selector, but the problem is we want to get only the nodes that correspond to the direct children of the given node.  In the case of something like `msubsup` (whose code is shared with `msub` and `msup`, when there is no `mjx-sup` element, we want to return `null`.  But a query selection might find an `mjx-sup` in an element further down (in the subscript or the base).  We want to avoid that.  Note that this version of the routine avoids walking into any child node that has an `id`.  Every element that comes from a MathML element will have an `id`, so we never walk into other MathML nodes.  That means we only check the structural elements that are part of the output for the parent node, not the renderings of any of the children.

Since the children could be arbitrarily complicated, query selector might have to look through *lots* of output.  Here, we are going to look through three or four elements at most.

zorkow:
Thanks for the explanation. Sounds reasonable for a special purpose algorithm. 
Can we put some of this into a comment for the function for future reference?

dpvc:
OK, I've added comments, and replaced the routine by the non-recursive version.

zorkow:
lgtm.

Issue #1370, opened 2016-02-03
[HTML-CSS] square root top line is too close to the number below


karoltarasiuk:
### Issue Summary

Square root top line is too close to the number below. I have also replicated the bug on latest develop branch using `unpacked/MathJax.js`.

### Steps to Reproduce

Simple go here: https://docs.staging.learnosity.com/demos/isolation/mathjax/260rendering.php

Or check this sample page:
```
<!DOCTYPE html>
<html>
<head>
<title>MathJax TeX Test Page</title>
<!-- Copyright (c) 2010-2015 The MathJax Consortium -->
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
<meta http-equiv="X-UA-Compatible" content="IE=edge" />
<meta name="viewport" content="width=device-width, initial-scale=1">

<script type="text/x-mathjax-config">
MathJax.Hub.Config({
    tex2jax: {inlineMath: [["$","$"],["\\(","\\)"]]}
});
</script>
<script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML-full"></script>

</head>
<body>

<p>
When $a \ne 0$, there are two solutions to \(ax^2 + bx + c = 0\) and they are
$$x = {-b \pm \sqrt{b^2-4ac} \over 2a}.$$
</p>
<p>
text around \(\frac{\sqrt{123}}{123}\) text around
</p>
<p>
$\frac{\sqrt{123}}{123}$
</p>

</body>
</html>
```

### Technical details

* MathJax Version: develop - latest commit: 0740dd4735cee63809c1228ae69bb2531e03c3fe
* Client OS: MaxOS
* Browser: Chrome 48

pkra:
Thanks for the detailed report. A few follow up questions.

* Since this is about a layout detail, could you add a screenshot to show what you see? 
* Which renderers (HTML-CSS, SVG, CommonHTML) are you seeing a problem with? I do see a minor issue with HTML-CSS.
* Your report points to the develop branch but your demo is using master. I'm guessing you see it on both?

karoltarasiuk:
1. A screenshot with the issue:
![screen shot 2016-02-04 at 9 12 03 am](https://cloud.githubusercontent.com/assets/1845230/12798792/8a02b708-cb1f-11e5-9e55-d2ccdab051bf.png)


The top line of a square root is too close to the number.

2. The issue is observed in HTML-CSS.
[...]

PR #1460, opened 2016-05-03
Adjust position of rule for square root when it is made via createRul…

dpvc:
Adjust position of rule for square root when it is made via `createRule()` in HTML-CSS output.  This improves the results for roots in fractions and super- or sub-scripts, or when the font is small.  Resolves issue #1370.


zorkow:
lgtm.

Issue #1441, opened 2016-04-15
[CommonHTML] problem with labeled equations


pkra:
Due to [this SO question](http://stackoverflow.com/questions/36628111/mathml-left-adjust-equation-numbers-in-table-of-equations), I noticed a problem with labels in the CommonHTML output.

The following example does have the labels float to the right/left (but does in HTML-CSS and SVG) -- but each table individually renders correctly.

```
<math display="block">
    <mtable side='left'>
        <mlabeledtr>
            <mtd>
                <mtext>(1)</mtext>
            </mtd>
            <mtd>
                <mi>X</mi>
            </mtd>
        </mlabeledtr>
    </mtable>
    <mspace linebreak="newline" />
    <mtable>
        <mlabeledtr>
            <mtd>
                <mtext>(2)</mtext>
            </mtd>
            <mtd>
                <mi>Y</mi>
            </mtd>
        </mlabeledtr>
    </mtable>
</math>
```

pkra:
For completeness, two labeled rows in one table are fine.

dpvc:
It has to do with some elements not getting `width:100%` when they should.  I'll see about a fix.

dpvc:
The `issue1441` branch of my fork of MathJax fixes the problem.

dpvc:
==> Merged.

dpvc:
==> In testsuite

MathMLToDisplay/Presentation/TablesAndMatrices/issue1441.html

PR #1462, opened 2016-05-04
Bubble percentage widths up while line breaking in CHTML.

dpvc:
This fixes the multiple-tagged-equationwith-line-breaks issue that Peter reported in #1441 by making sure the line box gets `width:100%` when the line contains an item with 100% width.

### DIFF
--- a/unpacked/jax/output/CommonHTML/autoload/multiline.js
+++ b/unpacked/jax/output/CommonHTML/autoload/multiline.js
@@ -348,6 +348,7 @@ MathJax.Hub.Register.StartupHook("CommonHTML Jax Ready",function () {
         //
         var node = this.CHTMLnodeElement();
         line.appendChild(node);
+        if (this.CHTML.pwidth && !line.style.width) line.style.width = this.CHTML.pwidth;
         //
         //  If it is last, remove right margin
         //  If it is first, remove left margin


unpacked/jax/output/CommonHTML/autoload/multiline.js
```diff
@@ -348,6 +348,7 @@ MathJax.Hub.Register.StartupHook("CommonHTML Jax Ready",function () {
         //
         var node = this.CHTMLnodeElement();
         line.appendChild(node);
+        if (this.CHTML.pwidth && !line.style.width) line.style.width = this.CHTML.pwidth;
```
zorkow: Do you also want to overwrite if line.style.width = 0?
Likewise if pwidth is 0 and width not set?

dpvc: `pwidth` is a percentage width (a string), so always has a percent (and as a string, `"0"` is not falsey).  Similarly, `line.style.width` will only be set (in this context) if it was set to a percent, and again, styles are always strings (even if set as numbers originally), so even width of 0 will be `"0"` and not false.  In most browsers, 0 is changed to `"0px"` in any case.

zorkow: Ah the pleasure of types...
So I assume none can be the empty string either?

dpvc: Yes, when `line.style.wodth` has not been set, it will be an empty string (not `null`).  That is the case we are testing for, here.  Once it is set, it will be non-empty (either a percentage, or something with an explicit `em` or `px` measurement).

zorkow:
lgtm.

Issue #1428, opened 2016-04-04
Progress bar keeps growing with dynamically created math


TomasRiker:
I have a case where math is dynamically created based on user input. Over time, the MathJax progress bar ("Typesetting ...") grows beyond all limits. I wish there was an (official) way of resetting it. I can reset `MathJax.Message.typesetting` etc. by hand, but it feels like MathJax should do this automatically after finishing its job.

dpvc:
Good point.  This is with `messageStyle` set to `simple`, I assume.  I agree that it should be reset for the next typeset operation.

dpvc:
The `issue1428` branch of my fork of MathJax fixes the issue.

dpvc:
==> Merged.

dpvc:
==> In testsuite

UI/issue1428.html

PR #1463, opened 2016-05-04
Reset message strings for messageStyle=simple for each typeset

dpvc:
This adds a `clearCounts()` function for the `MathJax.Message` object to clear the dot counts, and calls that each time a new typeset operation is performed so that the dots will not build forever if there are multiple typeset calls.  Resolves issue #1428


zorkow:
lgtm.

Issue #1419, opened 2016-03-22
ResetEquationNumbers


rylan:
Calling ```MathJax.Hub.Queue( ["resetEquationNumbers", MathJax.InputJax.TeX] );``` on a page without tex can cause Error("Can't make callback from given data") to be thrown intermittently.

dpvc:
By "without TeX" do you mean one that doesn't load the TeX input jax, or one with no math notation in the page?

You also haven't said where this call is coming in your page, which could also make a difference.  For example, since the resetEquationNumbers method is part of the AMSmath extension, if it came before that extension is loaded, then that could account for it.

You also don't say what your configuration is.  Are you loading a combined configuration file (via `?config=...` on the script that loads `MathJax.js`), or doing an in-line configuration.  I think all the TeX-based combined configuration files load the AMSmath extension, so that suggests you are doing a configuration by hand.  Do you load the AMSmath extension?  If so, is it in the TeX extensions array?

If it is in the TeX extensions array, that would mean that the extension isn't loaded until the main TeX jax is loaded, and the main TeX jax isn't loaded unless there is actual TeX notation on the page.

In this case, there are two possible solutions.  One would be to move the AMSmath extension to the main extensions array as in

```
MathJax.Hub.Config({
  extensions: ["TeX/AMSmath.js"]
});
```

which will force the extension to load at configuration time rather than during the typesetting phase, so the `resetEquationNumbers` would be available even when there is no TeX on the page.
 
Alternatively, you could use

```
if (MathJax.InputJax.TeX.resetEquationNumbers) MathJax.Hub.Queue(["resetEquationNumbers",MathJax.Hub]);
```

so that if the AMSmath extension hasn't run, you won't try to queue the call.

dpvc:
From MathJax's end, it would be possible to have a no-op `resetEquationNumbers` method in the `TeX/config.js` file that is overridden when the AMSmath extension is loaded, but I'm not sure that is really required, since testing for whether it exists is pretty straight-forward.

pkra:
@dpvc so this is `cannot reproduce`?

dpvc:
No, it can be reproduced under the right circumstances.  I just don't know which circumstances are the ones that pertain for this OP.  People using the combined configuration files that include the AMSmath extension (all the TeX) shouldn't experience this.

rylan:
[...]

PR #1464, opened 2016-05-04
Make sure resetEquationNumbers is always defined.

dpvc:
Add a dummy `resetEquationNumbers()` function that is replaced by AMS.  This makes sure the method is always available (when the TeX input jax is loaded), even if there is no TeX on the page, or AMSmath isn't loaded.  Resolves issue #1419.


zorkow:
lgtm.

Issue #1437, opened 2016-04-11
[non-SVG outputs] revisit role=math


pkra:
We are currently adding role=math to the output parent. However, this leads to poor results in the current AT landscape, cf. the test results at http://codepen.io/pkra/pen/zqrwQx and also https://github.com/w3c/aria/issues/290.

It seems better to remove role=math (since it breaks labels where present). If there's no suitable information in the source, we might want to consider adding a generic aria-label.

(For comparison, for the SVG output, we had changed the role to img https://github.com/mathjax/MathJax/issues/1192.)

dpvc:
> If there's no suitable information in the source, we might want to consider adding a generic aria-label.

Can you say more what you have in mind for this?

pkra:
> Can you say more what you have in mind for this?

I think I was thinking of MathML with alttext, TeX in semantics or an aria-label. But thinking about it now that's really a separate feature request. 

This issue was mainly about removing role=math since AT will ignore anything with that role -- even textContent is better than just silence, I think.

dpvc:
MathML's `alttext` is already used for the `aria-label`, if there isn't already one in pace, but we don't look for `<semantics>` content to use if there isn't an `alttext` or `aria-label`.  I think you are right that that is a different issue.

pkra:
> MathML's alttext is already used for the aria-label, if there isn't already one in pace,

Ah, I didn't know this -- thanks!

dpvc:
==> Merged.

dpvc:
==> In testsuite

MathMLToDisplay/issue1437.html

PR #1465, opened 2016-05-04
Don't set role in CHTML and HTML-CSS output

dpvc:
Don't set role in CHTML and HTML-CSS output, and fix comments in other jax.  Resolves issue #1437.


pkra:
LGTM.

zorkow:
+1

Issue #1474, opened 2016-05-15
[PreviewHTML] empty overset produces math processing error


dpvc:
When PreviewHTML (or fast-preview) is used, `\overset{}X` and `\unerset{}X` both generate [Math Processing Error].

Originally reported on [meta.math.SE](http://meta.math.stackexchange.com/questions/23215/math-processing-errors-on-empty-overset-and-differences-between-preview-and-a).

dpvc:
This is resolved in the `issue1474` branch of my fork of MathJax

dpvc:
==> Merged.

dpvc:
==> In testsuite

LaTeXToMathML/above-below/issue1474.html

PR #1475, opened 2016-05-15
Avoid error when \overset or \underset is empty.

dpvc:
This makes sure a child element is available before trying to modify it. Resolves issue #1474.


zorkow:
lgtm.

Issue #1478, opened 2016-05-17
Line breaking inside shrink-wrap elements doesn't work well


dpvc:
For elements that shrink-wrap their contents (like table cells or inline-block elements), the container width is measured as 0, and so line breaking can't be properly done.

Find a way to get the proper container width in that case (while still handling floating elements and other things that affect the width).

See issue #571, and also #1466.

dpvc:
The `issue1478` branch of my fork of MathJax includes changes that gets the proper container width for math inside shrink-wrapped elements.  It also makes all three output jax use the same mechanism for determining the container width.

dpvc:
==> Merged.

dpvc:
==> Merged

dpvc:
==> In testsuite

MathMLToDisplay/issue1478.html

PR #1479, opened 2016-05-17
Properly determine container widths.  Resolves issue #1478.

dpvc:
This path changes how the container width is determined so that it can be used within shrink-wrap elements like table cells and inline-block elements, while still properly handling floating elements that reduce the width.

The old approach was to use an `<hr>` element to get the width, since that would fill the available space, but not overlap floating elements.  The problem with that approach is that shrink-wrap elements would end up being width 0 if they don't have any other content than the math.

To resolve this, we can use an element with `display:table-cell` and `width` set to some large amount, since the table will shrink the cell to the available space, but the large width will force it to be as large as possible. Thus it gets the full width of that is available.  (This was the technique used in the `handle-floats` extension for displaying the HTML-CSS output within the available space, but it was not used for measuring the space.)  This is more effective than the `<hr>` technique, since it means that in a shrink-wrapped element, the table cell will try to be as large as possible (forcing the container to be as large as possible), whereas the rule would not, leaving the width of 0.

The HTML-CSS and CHTML output now include the CSS used in the `handle-floats` extension to make sure that the output fits within the floating elements.  This was not being done with the CHTML output (and there was no `handle-floats` extension for it), and this makes the current `handle-floats` for HTML-CSS obsolete.

I would recommend reviewing the SVG output first, then the HTML-CSS and finally the CHTML output.

**Note**: IE < 8 will not get the container width correct inside shrink-wrap elements, since it doesn't implement `display:table-cell` properly, and we are forced to use `display:block; width:100%` instead, which doesn't force the container to be as wide as possible.  But since `display:block` is also implemented incorrectly in IE < 8, it does get the proper width between floating elements when not used inside a shrink-wrap element.  This only applies to HTML-CSS output, since SVG only works in IE9 and above, and CHTML only in IE8 and above.

### DIFF
--- a/unpacked/jax/output/SVG/jax.js
+++ b/unpacked/jax/output/SVG/jax.js
@@ -2142,7 +2142,7 @@
             if (shift) {
               HUB.Insert(style,({
                 left: {marginLeft: SVG.Ex(shift)},
-                right: {marginRight: SVG.Ex(-shift), marginLeft: SVG.Ex(Math.max(0,shift-(w)))},
+                right: {marginRight: SVG.Ex(-shift), marginLeft: SVG.Ex(Math.max(0,shift-w))},
                 center: {marginLeft: SVG.Ex(shift), marginRight: SVG.Ex(-shift)}
               })[values.indentalign]);
             }
--- a/unpacked/jax/output/SVG/jax.js
+++ b/unpacked/jax/output/SVG/jax.js
@@ -100,6 +100,12 @@
           "min-height": 0, "max-height":"none",
           padding:0, border: 0, margin: 0
         },
+        ".MathJax_SVG_LineBox": {
+          display: "table-cell!important",
+          width: "10000em!important",
+          "min-width":0, "max-width":"none",
+          padding:0, border:0, margin:0
+        },
         
         "#MathJax_SVG_Tooltip": {
           position: "absolute", left: 0, top: 0,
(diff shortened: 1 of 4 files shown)


unpacked/jax/output/SVG/jax.js
```diff
@@ -2131,7 +2142,7 @@
             if (shift) {
               HUB.Insert(style,({
                 left: {marginLeft: SVG.Ex(shift)},
-                right: {marginRight: SVG.Ex(-shift), marginLeft: SVG.Ex(Math.max(0,shift-(l+svg.w+r)))},
+                right: {marginRight: SVG.Ex(-shift), marginLeft: SVG.Ex(Math.max(0,shift-(w)))},
```
zorkow: Is there something missing? O/w maybe remove parentheses around (w).

zorkow:
I can't really comment much on the code. But if it works, fine.

unpacked/jax/output/SVG/jax.js
zorkow: I get it. It's a query-replace artifact.

dpvc: Right.  I'll fix it.

dpvc: Fixed.

zorkow:
lgtm.

Issue #1477, opened 2016-05-17
AssistiveMML extension throws error on IE8 [was: HTML-CSS output causes “unknown runtime error” in IE8]


drify:
Just open the test page http://cdn.mathjax.org/mathjax/latest/test/sample-tex.html using IE8. In fact the math renders properly, but the console shows the error. 
MathJax 2.5 doesn't have this issue. http://cdn.mathjax.org/mathjax/2.5-latest/test/sample-tex.html

pkra:
Thanks for the report.

The error traces back to https://github.com/mathjax/MathJax/blob/master/unpacked/extensions/AssistiveMML.js#L131 which might be a MathML parsing problem with IE8 -- though I'm a little confused why this wouldn't occur with the CommonHTML as well. 

Workaround is to disable the AssitiveMML extension; it's probably useless on IE8 anyway since no AT needs/makes use of it on IE8, I think (only MathPlayer is relevant and that works differently after all).

dpvc:
I put a `try {...} catch (err) {}` around that line, which will prevent the error message, but doesn't solve the problem that is preventing the MathML from being inserted.  I don't think it is worth trying to find that, since (as you say), there is no AT that works with it.

The `issue1477` branch of my fork of MathJax includes the patch.  I think we can include this in 2.x.x, though I don't really have strong feelings either way.

pkra:
>  I don't think it is worth trying to find that, since (as you say), there is no AT that works with it.

:+1: 

>  I think we can include this in 2.x.x, though I don't really have strong feelings either way.

:+1:

dpvc:
==> Merged.

PR #1483, opened 2016-05-19
Catch IE8 errors with inserting MathML from AssistiveMML extension. #1477

dpvc:
Since there is no reader that can handle this with IE8 anyway, there isn't much point in trying to figure out whatever bug is the source of this error.  Resolves issue #1477.


pkra:
LGTM.

Issue #1449, opened 2016-04-26
support linebreaks in munder


pkra:
It would be nice if something like the following would work. 

```html
<math xmlns="http://www.w3.org/1998/Math/MathML">
  <munder>
    <mrow class="MJX-TeXAtom-OP MJX-fixedlimits">
      <munder>
        <mrow>
          <mi>f</mi>
          <mo stretchy="false">(</mo>
          <mi>t</mi>
          <mo>+</mo>
          <mn>1</mn>
          <mo stretchy="false">)</mo>
        </mrow>
        <mo>&#x23DF;<!-- ⏟ --></mo>
      </munder>
    </mrow>
    <mrow class="MJX-TeXAtom-ORD">
      <mtext>distance at&#xA0;</mtext>
      <mo linebreak="newline"></mo>
      <mtext>time&#xA0;</mtext>
      <mi>t</mi>
      <mo>+</mo>
      <mn>1</mn>
    </mrow>
  </munder>
</math>
```

dpvc:
If you remove the `class="MJX-TeXAtom-ORD"` from the mrow surrounding the line break, it will work.  It turns out that the TeXAtom node is not considered a container whose contents can be broken.  Perhaps that needs to be changed.  You may also wish to add `indentalign="center"` to the mo that is used to obtain the newline.

pkra:
Oh, nice! I shouldn't have been so lazy (constructing this example in TeX) ;-).

dpvc:
The `issue1449` branch of my fork of MathJax includes a fix for this.

dpvc:
==> Merged.

dpvc:
==> In testsuite

MathMLToDisplay/Topics/LineBreak/issue1449.html

PR #1485, opened 2016-05-19
Allow line breaks within TeXAtom elements  Resolves issue #1449.

dpvc:
Since TeXAtoms are basically mrows with additional properties, it makes sense to allow line breaks.


pkra:
LGTM.

Issue #1499, opened 2016-06-02
[CommonHTML] semantic tag causes shrink-wrapping of labels


pkra:
The following gets shrink-wrapped:

```html
<math>
  <semantics>
    <mtable>
      <mlabeledtr>
        <mtd><mtext>(1)</mtext></mtd>
        <mtd><mtext>Hello</mtext></mtd>
      </mlabeledtr>
    </mtable>
  </semantics>
</math>
```

If the semantics tag is removed, the rendering is correct.

dpvc:
This is resolved in the `issue1499` branch of my fork of MathJax.

dpvc:
==> Merged.

dpvc:
==> In testsuite

MathMLToDisplay/issue1499.html

PR #1501, opened 2016-06-02
Allow percentage width to bubble up in bbox.updateFrom(cbox). #1499

dpvc:
This PR resolves issue #1499, where `<semantics>` would shrink-wrap an `<mtable>` with 100% width (e.g., one with an `<mlabeledtr>`).


zorkow:
lgtm.

Issue #1500, opened 2016-06-02
[CommonHTML] equation label not properly aligned in large arrays with many labels


pkra:
Examples

* Tags become vertically lower than the cell

```latex
 \begin{eqnarray} 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
 a \tag{tag}\\ 
  \end{eqnarray}
```

* Tags become vertically higher than the cell

```latex
\begin{eqnarray} 
&& a \tag{tag}\\ 
&& a \tag{tag}\\ 
&& a \tag{tag}\\ 
&& a \tag{tag}\\ 
&& a \tag{tag}\\ 
&& a \tag{tag}\\ 
&& a \tag{tag}\\ 
&& a \tag{tag}\\ 
 \end{eqnarray}
```

Screenshot (combined)

![labels](https://cloud.githubusercontent.com/assets/426686/15747209/f6ecc8aa-28d9-11e6-9099-d05302adb2e2.png)

pkra:
Oh wow. 1500 bugs!

dpvc:
It turns out that there were several issues that contributed to this.  One was that the heights of the lines were being adjusted to the wrong value (argh).  But there were also issues with empty cells (it turns out inline-block node with no text content align differently from those with text content), and with nodes whose height is less than the midpoint of the font (didn't appear in your examples, but did in my own tests).

I think I have taken care of all of them in the `issue1500` branch of my fork of MathJax.

dpvc:
==> Merged.

dpvc:
> Oh wow. 1500 bugs!

Well, because pull requests use the same numbers as issues (and we've been using more pull requests lately), there aren't really 1500 bugs.  There are 1220 issues (not all of them bugs) and 204 pull requests as of this posting.  So I'm not sure what the other items 76 are!

pkra:
> So I'm not sure what the other items 76 are!

Hm... the numbers I see do add up (maybe you had a filter on?) But yes, over 200 PRs in total -- still awesome :-)

dpvc:
==> In testsuite

MathMLToDisplay/Presentation/TablesAndMatrices/issue1532.html

PR #1503, opened 2016-06-06
Handle adjusting cell heights properly. #1500

dpvc:
The actual height and depth of the row is in `state.H[i]` and `state.D[i]`, but we were using `LH` and `LD`, the usual line height and depth.  This PR fixes that.  This works if there is actual text in the cell, but if the cell is empty, the margin-bottom doesn't fall below the baseline, it sits on top of the baseline, which spoils the alignment.  Also, if the height of the cell is less than the midpoint of the font, the browser adds extra height making it the midpoint of the font, and we didn't compensate for that.

Both issues are resolved by adding a 1em strut to the cell (so it has content and is not too short), and compensating for that in the height adjustments.  (Since the font in question was the surrounding font, it was not easy to tell where its midpoint was; this solution is independent of the surrounding font.)

Resolves issue #1500.


zorkow:
lgtm.

pkra:
@dpvc I have trouble with this in mathjax-node (using the develop branches of both repos); the result keeps having this problem. Does that sound like it could be another jsdom bug or should I better triple check my setup? :frowning:

PR #1504, opened 2016-06-06
Make getJaxForMath() work even during chunking.

dpvc:
Make `getJaxForMath()` work even during chunking in all output modes.  (The chunking code adds `MathJax_Processed` to the class, so we need to search for the class we need, not check if the class is equal to it.)

Resolves issue mathjax/MathJax-RespEq#119.


zorkow:
lgtm.

PR #1507, opened 2016-06-07
Don't add a 'src' attribute (<image> uses href, not src)

dpvc:
The `src` attribute is no valid on an `<image>` tag in SVG.  It is the `href` attribute (added in the `BBOX.MGLYPH()` function) that actually does the linking to the image source.


zorkow:
lgtm.

Issue #1023, opened 2015-01-27
\middle preceded by primed symbol gives an error.


nberth:
The following code gives me an error "\middle must be within \left and \right" when I render it in my browser (Firefox 21.0, MathJax v2.4.0):

- `\left\{q\in Q \middle| P(q)\right\}` works;
- `\left\{q\in Q' \middle| P(q)\right\}` doesn't, nor `\left\{q\in Q'' \middle| P(q)\right\}` incidentally.
This also happens when `Q'` is the result of macro expansion:
- `\newcommand{\Q}{Q} \left\{q\in \Q \middle| P(q)\right\}` works;
- `\newcommand{\Q}{Q'} \left\{q\in \Q \middle| P(q)\right\}` doesn't.

The error is reproducible (though no error message is displayed) on the dynamic rendering page [1], yet I did not try it on the development branch for lack of time to install it (can anyone try?).

[1] http://cdn.mathjax.org/mathjax/latest/test/sample-dynamic.html

pkra:
Thank you for reporting the issue! (I've modified your posting to escape the TeX). We'll look into it.

dpvc:
For now you can use 

    \left\{q\in {Q'} \middle| P(q)\right\}

to get the result you are looking for.  There is a problem with how `\middle` detects whether it is within `\left...\right` (obviously).

dpvc:
The `issue1023` branch of my fork of MathJax resolves this issue (as well as a potential spacing problem with `\middle`).

dpvc:
==> Merged.

dpvc:
==> In testsuite

LaTeXToMathML/delimiters/issue1023.html

PR #1508, opened 2016-06-07
Fix problems with \middle.  #1023

dpvc:
Surround \middle with OPEN and CLOSE TeXAtoms, so that the spacing will be consistent with TeX's version (e.g., \mathrel{}\middle|\mathrel{} will produce the right spacing).  As a side effect, this resolves the problem with the quotes, sot this resolves issue #1023.

### DIFF
--- a/unpacked/jax/input/TeX/jax.js
+++ b/unpacked/jax/input/TeX/jax.js
@@ -1335,9 +1335,11 @@
     
     Middle: function (name) {
       var delim = this.GetDelimiter(name);
+      this.Push(MML.TeXAtom().With({texClass:MML.TEXCLASS.CLOSE}));
       if (this.stack.Top().type !== "left")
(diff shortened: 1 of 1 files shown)


zorkow:
lgtm.

Question: What happens on a TEX.ERROR? Will that lead to a math processing error?

dpvc:
> What happens on a TEX.Error? 

The TeX input processor will stop parsing the expression and produce an `merror` element that includes the error message.  If the `noErrors` extension is in play, then the error causes the original TeX code to be displayed rather than the error message (this was requested by AMS long ago, though personally I think it is a mistake to have it in the combined config files).

> Will that lead to a math processing error?

No.  Those should only come from errors in the MathJax javascript code, not from errors in the user input (which should be reported like the one at line 1340 here).

pkra:
> though personally I think it is a mistake to have it in the combined config files).

+1

Issue #1084, opened 2015-03-08
Arrows are not symmetric with \left...\right


dpvc:
Vertical arrows are not marked as symmetric when used with `\left` and `\right`, but should be.

dpvc:
This is resolved in the `issue1084` branch of my fork of MathJax.

dpvc:
==> Merged

dpvc:
==> In testsuite

LaTeXToMathML/delimiters/issue1084.html

PR #1509, opened 2016-06-07
Make delimiters symmetric for \left and \right.  #1084

dpvc:
Mark delimters are symmetric when used with `\left` and `\right`.  In particular, this makes the arrow delimiters be symmetric.  Resolves issue #1084.

### DIFF
--- a/unpacked/jax/input/TeX/jax.js
+++ b/unpacked/jax/input/TeX/jax.js
@@ -2210,9 +2210,9 @@
      */
     fenced: function (open,mml,close) {
       var mrow = MML.mrow().With({open:open, close:close, texClass:MML.TEXCLASS.INNER});
-      mrow.Append(MML.mo(open).With({fence:true, stretchy:true, texClass:MML.TEXCLASS.OPEN}));
+      mrow.Append(MML.mo(open).With({fence:true, stretchy:true, symmetric:true, texClass:MML.TEXCLASS.OPEN}));
       if (mml.type === "mrow") {mrow.Append.apply(mrow,mml.data)} else {mrow.Append(mml)}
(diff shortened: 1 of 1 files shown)


unpacked/jax/input/TeX/jax.js
```diff
@@ -2210,9 +2210,9 @@
      */
     fenced: function (open,mml,close) {
       var mrow = MML.mrow().With({open:open, close:close, texClass:MML.TEXCLASS.INNER});
-      mrow.Append(MML.mo(open).With({fence:true, stretchy:true, texClass:MML.TEXCLASS.OPEN}));
+      mrow.Append(MML.mo(open).With({fence:true, stretchy:true, symmetric:true, texClass:MML.TEXCLASS.OPEN}));
       if (mml.type === "mrow") {mrow.Append.apply(mrow,mml.data)} else {mrow.Append(mml)}
-      mrow.Append(MML.mo(close).With({fence:true, stretchy:true, texClass:MML.TEXCLASS.CLOSE}));
+      mrow.Append(MML.mo(close).With({fence:true, stretchy:true, symmetric:true, texClass:MML.TEXCLASS.CLOSE}));
```
zorkow: Just for clarification: Does that also mean that \left( \right] would be marked as symmetric?

dpvc: Yes, but that should not be a problem because: (a) they default to symmetric in the operator dictionary already (it was only the arrows that didn't), and (b) in TeX `\left` and `\right` always produce symmetric delimiters.  Note that symmetric will not appear in the `toMathML()` output (so not in the Show Math As... output) if the delimiter has symmetric as the default.

zorkow:
lgtm.

Issue #1356, opened 2016-01-22
Fences with separators="false" (true) causing processing errors


pkra:
As per @dpvc, they get converted to booleans, leading to errors.

dpvc:
The `issue1356` branch of my fork of MathJax resolves this issue.

dpvc:
==> Merged.

dpvc:
==> In testsuite

MathMLToDisplay/issue1356.html

PR #1511, opened 2016-06-07
Don't convert attribute to boolean unless the default is a boolean.  #1356

dpvc:
Don't convert attribute to boolean unless the default is a boolean.  Resolves issue #1356.

### DIFF
--- a/unpacked/jax/input/MathML/jax.js
+++ b/unpacked/jax/input/MathML/jax.js
@@ -142,8 +142,14 @@
         value = this.filterAttribute(name,value);
         var defaults = (mml.type === "mstyle" ? MML.math.prototype.defaults : mml.defaults);
         if (value != null) {
-          if (value.toLowerCase() === "true") {value = true}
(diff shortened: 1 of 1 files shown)


zorkow:
I don't quite understand what goes on in the code as the condition of the if is really a mystery to me. 
But if it works that's fine.

dpvc:
Yes, the condition is a bit complicated.  It is trying to convert a string to a boolean only when the type of the attribute is boolean, but we don't actually have any data on what the type should be.  So we look at the default value (if any) and if it is boolean, then convert, or if it is `"inherit"` (some booleans are inherited from parent elements, so this is really a punt), or if it is `"auto"` (a special internal value that means the value is computed automatically in some when when not given) and the default value for that computation is boolean, then convert.  All this is an attempt to determine whether the attribute should be boolean or not.

In v3.0, we may want to have a more complete description of valid MathML, and include that sort of information for this kind of use (as well as validating MathML input, which we don't currently do).

PR #1512, opened 2016-06-07
Make CHTML mglyph scale image size by hand.  mathjax/MathJax-node#241

dpvc:
Make CHTML `mglyph` scale image size by hand rather than using `offsetWidth` and `offsetHeight`, since those cause reflow. This also makes it work in mathjax-node.  Resolves mathjax/Mathjax-node#241.

### DIFF
--- a/unpacked/jax/output/CommonHTML/autoload/mglyph.js
+++ b/unpacked/jax/output/CommonHTML/autoload/mglyph.js
@@ -60,15 +60,15 @@ MathJax.Hub.Register.StartupHook("CommonHTML Jax Ready",function () {
           var img = CHTML.addElement(node,"img",{
             isMathJax:true, src:values.src, alt:values.alt, title:values.alt
           });
-          var w = bbox.img.img.width/CHTML.em, h = bbox.img.img.height/CHTML.em;
-          if (values.width !== "")  img.style.width  = CHTML.Em(this.CHTMLlength2em(values.width,w));
-          if (values.height !== "") img.style.height = CHTML.Em(this.CHTMLlength2em(values.height,h));
-          //
-          //  Warning:  causes page reflows
-          //
-          bbox.w = bbox.r = img.offsetWidth/CHTML.em; bbox.h = bbox.t = img.offsetHeight/CHTML.em;
+          var w = values.width, h = values.height;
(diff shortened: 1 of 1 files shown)


unpacked/jax/output/CommonHTML/autoload/mglyph.js
```diff
@@ -60,15 +60,15 @@ MathJax.Hub.Register.StartupHook("CommonHTML Jax Ready",function () {
...
-          //
-          //  Warning:  causes page reflows
-          //
-          bbox.w = bbox.r = img.offsetWidth/CHTML.em; bbox.h = bbox.t = img.offsetHeight/CHTML.em;
+          var w = values.width, h = values.height;
+          var W = bbox.img.img.width/CHTML.em, H = bbox.img.img.height/CHTML.em;
+          var WW = W, HH = H;
+          if (w !== "") {W = this.CHTMLlength2em(w,WW); H = (WW ? W/WW * HH : 0)}
+          if (h !== "") {H = this.CHTMLlength2em(h,HH); if (w === "") W = (HH ? H/HH * WW : 0)}
+          img.style.width  = CHTML.Em(W); bbox.w = bbox.r = W;
```
zorkow: So these will fire, even if w,h are undefined? Is that intended?

zorkow: For future reference: The code would be a lot easier to read if it were spread out a bit and had more descriptive var names. In particular cap vs non-cap is difficult to parse.

dpvc: > So these will fire, even if w,h are undefined? 

I'm not sure which line "these" refer to.  Note that `w` and `h` are never undefined, though they can be empty (they come from the `getValues()` method, and since the defaults for these are `""`, if there is no explicit value, those will eventually be found, so they are always defined).

If you mean the style settings in line 68, these use the values of `W` and `H`.  These come from the `img` width and height, and since the image has loaded when these lines run, they are always defined.  Even in jsdom, they are defined (but 0, so in mathjax-node, you are expected to use explicit width and height).

Does that cover your concerns?

zorkow:
Thanks for the explanation.
lgtm.

Issue #1497, opened 2016-05-31
PlainSource output vs AssistiveMML


pkra:
A user pointed out to me that it would be good for the PlainSource output to disable the AssitiveMML extension. 

I think that would be preferable behavior as it avoids both accessibility issues as well as copy&paste problems on some browsers when PlainSource is used.

dpvc:
The `issue1497` branch of my fork of MathJax will prevent assistive MathML output when PlainSource is the selected output.

pkra:
I guess that's a yes to my question :wink: . Sorry for not proposing a fix alongside it.

dpvc:
For copy&paste, it might be good for PlainSource to insert the math delimiters for TeX code.  Right now, it is just the TeX code directly, which may be good for reading, but not so good for copy and paste.  Should there be a menu option to control that?

pkra:
Yes, I thought about that as well. But then I coudn't decide if "standard" delims are better than page-configured ones so I left it.

dpvc:
Well, the tex2jax could store the actual delimiters used for each expression, so they would be available for PlainSource to use.

camlorn:
As someone who will use this feature, `$` isn't bad and `\(\)` is horrible.  You don't see the latter often, but it does come up every once in a while.

No delimiters is ideal for use with a screen reader, however.  I think this is probably the primary use case of this feature by far, and is certainly the reason it was originally suggested.  I can see the potential use to sighted people though.

dpvc:
@camlorn, thanks for your comment.

I agree that the screen-reader use case is best with no delimiters.  The other use case, however, is to be able to copy-and-paste paragraphs of math into an actual TeX document.  For that use case, having delimiters would be useful.  Because there are two distinct use cases with different needs (in terms of delimiters), it might be useful to allow the user to select whether he or she wants them or not.

pkra:
> The other use case, however, is to be able to copy-and-paste paragraphs of math into an actual TeX document.

Right. Then again, we also don't use delims in the show-math popups.

dpvc:
> Then again, we also don't use delims in the show-math popups.

Of course, there is nothing else to delimit it from on the pop up.  Here, the idea is to be able to copy and paste entire paragraphs or longer, with multiple equations and intervening text.  There, delimiters are critical.

pkra:
[...]

PR #1513, opened 2016-06-07
Don't do assistive MathML when the output renderer is PlainSource.  #1497

dpvc:
Don't do assistive MathML when the output renderer is PlainSource.  Resolves issue #1497.


pkra:
LGTM.

Issue #1415, opened 2016-03-15
[TeX input] \class is lost during input [was "\class" being ignored in "\left[\class{test}{\sin(x)}\right]"]


TomasRiker:
In `\left[\class{test}{\sin(x)}\right]`, the `\class` is ignored.

When replacing `\sin(x)` with e.g. `x^2`, it works. It also works when removing the `\left[` and `\right]` around the expression.

See this screenshot (the `test` class is just defined to be `color: red`):

![mathjax bug](https://cloud.githubusercontent.com/assets/295205/13771996/f67e9f5a-ea91-11e5-80d5-f2b3ec8944fb.png)

pkra:
Thanks for the report. I can confirm this.

It looks like the TeX input processor is losing the class.

dpvc:
This turns out to be due to [this line](http://github.com/mathjax/MathJax/blob/master/unpacked/jax/input/TeX/jax.js#L2214) in the TeX input jax, which unwraps an `row` when it adds fences to it.  In this case, the class was attached to the `mrow`.  The code should not unwrap the `mrow`, but jest leave it as it is, even though most of the time it will be unnecessary.

dpvc:
The `issue1415` branch of my fork of MathJax resolves this issue.

dpvc:
==> Merged.

dpvc:
==> In testsuite

LaTeXToMathML/delimiters/issue1415.html

PR #1514, opened 2016-06-07
Don't unwrap mrow when created fenced elements.  #1415

dpvc:
Don't unwrap mrow when created fenced elements (in case attributes have been attached to the mrow).  Resolves issue #1415.


zorkow:
lgtm.

Issue #1519, opened 2016-06-14
Add [Contrib] to the default paths configuration


dpvc:
This will make it easier for people to load the items from the third-party repository.

dpvc:
The `issue1519` branch of my fork of MathJax includes this.

dpvc:
==> Merged.

dpvc:
==> In testsuite

Configuration/Contrib.html

PR #1522, opened 2016-06-14
Add Contrib to the Ajax paths. #1519

dpvc:
Add `[Contrib]` to the Ajax paths.  Resolves issue #1519.

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -659,6 +659,8 @@ MathJax.cdnFileVersions = {};  // can be used to specify revisions for individua
   
   var PATH = {};
   PATH[BASENAME] = "";  // empty path gets the root URL
+  PATH.Contrib = (String(location.protocal).match(/^https?:/) ? "" : "http:") +
+                   "//cdn.mathjax.org/mathjax/contrib";   // the third-party extensions
   
   BASE.Ajax = {
     loaded: {},         // files already loaded


pkra:
LGTM.

Should we consider enforcing `https` in v3.0? Right now, this would break very old systems like WinXP with outdated cert infrastructures but perhaps it is time for 3.0.

dpvc:
Is there anything really that needs to be done differently for `https` from `http`?  If the only accommodation that needs to be made is to check the protocol now and then, I don't see the harm in doing so.  But I don't have any strong feelings about it.

pkra:
> If the only accommodation that needs to be made is to check the protocol now and then, I don't see the harm in doing so.

Right. I suppose if the CDN always loaded via https, then at least it can't be a point of attack. Which, arguably, is very little if the site is on http.

> But I don't have any strong feelings about it.

Me neither.

But now I spotted a typo!

unpacked/MathJax.js
```diff
@@ -659,6 +659,8 @@ MathJax.cdnFileVersions = {};  // can be used to specify revisions for individua
   
   var PATH = {};
   PATH[BASENAME] = "";  // empty path gets the root URL
+  PATH.Contrib = (String(location.protocal).match(/^https?:/) ? "" : "http:") +
```
pkra: `location.protocol`?

pkra: Sorry for not catching that before.

dpvc: Argh!  I'll fix it.

Issue #1468, opened 2016-05-12
Arrays should reset font selection in TeX


dpvc:
Currently, `\mathbf{\begin{array}{cc}x&y\end{array}}` produces a bold `x` (and a normal `y`), whereas in actual LaTeX, the `x` is normal not bold.

dpvc:
The `issue1468` branch of my fork of MathJax includes the fix for this.

dpvc:
==> Merged.

dpvc:
==> In testsuite

LaTeXToMathML/matrices/issue1468.html

PR #1523, opened 2016-06-14
Don't copy environment into array environments.  #1468

dpvc:
Don't copy environment into array environments.  Use

```
\mathbf{\begin{array}{cc}x&y\end{array}}
```

to test it (the `x` should not be in bold).  Resolves issue #1468.


zorkow:
lgtm.

Issue #1445, opened 2016-04-17
The tex2jax and other preprocessors should not make a preview if there already is one


dpvc:
Currently, tex2jax and the other previewers will make a preview even if there is already an element with `class="MathJax_Preview"`.  They probably should skip the preview in that case.

dpvc:
The `issue1445` branch of my fork of MathJax fixes this.

dpvc:
==> Merged.

dpvc:
==> In testsuite

Parsing/issue1445.html

PR #1524, opened 2016-06-14
Don't make a preview if there already is one.  #1445

dpvc:
Don't have preprocessors make a preview if there already is one.  Resolves issue #1445.


zorkow:
lgtm.

Issue #1403, opened 2016-03-01
Make the MathJax root configurable


dpvc:
Currently, using 

    MathJax.Hub.Config({root: '...'});

doesn't cause the root to be copied to `Ajax.root`.  There should be a point at the end of the configuration process, but before files are loaded, where the `root` is copied to `Ajax.root`.  This may be complicated, because the combined configuration file is loaded before user configuration is processed, so that may need to be taken into account.

dpvc:
The `issue1403` branch of my fork of MathJax includes a fix for this.

dpvc:
==> Merged

dpvc:
==> In testsuite

Configuration/Hub/root.html

alexweissman:
This really needs to be added to the documentation.

pkra:
Good point => https://github.com/mathjax/MathJax-docs/issues/182

PR #1525, opened 2016-06-14
Allow MathJax root to be configured.  #1403

dpvc:
Allow MathJax root to be configured.  Resolves issue #1403.

The best way to set the root is to use

```
<script>
MathJax = {root:"http://myhost.net/mathjax"};
</script>
<script src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML"></script>
```

so that the configuration file will be loaded from the correct location as well. Otherwise, you can use

```
MathJax.Hub.Config({
  root:"http://myhost.net/mathjax",
  config: ["TeX-AMS_HTML.js"]
});
</script>
<script src="http://cdn.mathjax.org/mathjax/latest/MathJax.js"></script>
```

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -2503,6 +2503,11 @@ MathJax.Hub.Startup = {
   Config: function () {
     this.queue.Push(["Post",this.signal,"Begin Config"]);
     //
+    //  Make sure root is set before loading any files
+    //
+    if (MathJax.AuthorConfig && MathJax.AuthorConfig.root)
+      MathJax.Ajax.config.root = MathJax.AuthorConfig.root;
+    //
     //  If a locale is given as a parameter,
     //    set the locale and the default menu value for the locale
     //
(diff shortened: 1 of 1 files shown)


unpacked/MathJax.js
```diff
@@ -2552,15 +2557,15 @@ MathJax.Hub.Startup = {
...
     for (var i = 0, m = scripts.length; i < m; i++) {
       var type = String(scripts[i].type).replace(/ /g,"");
       if (type.match(/^text\/x-mathjax-config(;.*)?$/) && !type.match(/;executed=true/)) {
         scripts[i].type += ";executed=true";
-        last = queue.Push(scripts[i].innerHTML+";\n1;");
+        queue.Push(scripts[i].innerHTML+";\n1;");
       }
     }
-    return last;
+    return queue.Push(function () {MathJax.Ajax.config.root = MathJax.Hub.config.root});
```
zorkow: Is Push synchronous, i.e., will there be a return value?

dpvc: Yes, the return value is the `MathJax.Callback` object created from the item pushed (callback can be created from lots of possible data).  The important point in returning the last callback pushed is that `Callback` objects are not only function that can be called, but they also know how to work with `MathJax.Callback.Queue` objects; when a `Callback` is pushed on a `Queue`, the `Queue` will block until the `Callback` is called, and when it is, the next item in the `Queue` will start.  But if a queued item returns another `Callback`, then the queue continues to pause until *that* callback is called, and so on.

So in this case, `ConfigBlocks` is itself pushed on a queue, and that queue will not continue until all the items that have been queued in `queue` in this routine are run, because the outer queue will have to wait for the `Callback` object returned by `queue.Push()`, and that only runs after everything else in `queue` has been run.

The `MathJax.Callback` was my attempt to do the job that now would be handled by promises, but that idea hadn't been developed (in javascript) at the time, and so the paradigm is a bit more awkward than it should be.  It is one of the things that needs fixing in 3.0.

zorkow:
Thanks for the explanation. 
lgtm.

Issue #989, opened 2015-01-03
Negative sign in <mn> doesn't use the math minus sign


dpvc:
The HTML-CSS and SVG output display `<mn>-1</mn>` using the dash U+002D rather than the minus sign U+2212.  It would be nice to convert that to the correct character.

Firefox's native MathML rendering also uses U+002D for `<mn>-1</mn>`.  Should we tweak that as well?

davidcarlisle:
On 3 January 2015 at 16:59, Davide P. Cervone <notifications@github.com>
wrote:

> The HTML-CSS and SVG output display <mn>-1</mn> using the dash U+002D
> rather than the minus sign U+2212. It would be nice to convert that to the
> correct character.
>
> Firefox's native MathML rendering also uses U+002D for <mn>-1</mn>.
> Should we tweak that as well?
>

we added specific notes to the mathml spec telling people to code negative
numbers as
<mrow><mo>-</mo><mn>123</mn></mrow>
but it's not at all uncommon to see <mn>-123</mn> so making it use &minus;
of it doesn't contort the code to much is probably the pragmatic thing to
do, although I think 2D is "more correct mathnml rendering" it's almost
certainly not what the author intended.

David

dpvc:
Thanks, David.  The issue actually car up because the new content-mathml output produces `<mn>-1</mn>` rather than `<mrow><mo>-</mo><mn>1</mn></mrow>`.  That should probably be fixed, but I thought it would be good to handle `<mn>-1</mn>` as though it were the recommended form.

davidcarlisle:
On 3 January 2015 at 21:16, Davide P. Cervone <notifications@github.com>
wrote:

> Thanks, David. The issue actually car up because the new content-mathml
> output produces <mn>-1</mn> rather than <mrow><mo>-</mo><mn>1</mn></mrow>.
> That should probably be fixed, but I thought it would be good to handle
> <mn>-1</mn> as though it were the recommended form.
>
>

oops I should definitely fix that!

(I can be relaxed about other people not following the spec but me not
following the spec doesn't look too good:-)

David

dpvc:
Since Christian made the version of your code that we are using, it may be that we need to do the change.  Can you suggest where that location might be?  I haven't looked closely at the code, but will go through it if necessary.

davidcarlisle:
On 3 January 2015 at 21:36, Davide P. Cervone <notifications@github.com>
wrote:

> Since Christian made the version of your code that we are using, it may be
[...]

PR #1526, opened 2016-06-15
Make minus in <mn> produce U+2212 rather than U+002D.  #989

dpvc:
Make minus in `<mn>` produce U+2212 rather than U+002D.  Resolve issue #989.

This also tweaks the NativeMML output; should we do that?  Or make a configuration parameter to prevent that?

### DIFF
--- a/unpacked/jax/output/SVG/jax.js
+++ b/unpacked/jax/output/SVG/jax.js
@@ -1566,6 +1566,31 @@
 	return svg;
       }
     });
+    
+    MML.mn.Augment({
+      SVGremapMinus: function (text) {return text.replace(/^-/,"\u2212")},
+      toSVG: function () {
+        this.SVGgetStyles();
+	var variant = this.SVGgetVariant();
+        var svg = this.SVG(); this.SVGgetScale(svg);
+        this.SVGhandleSpace(svg);
+        var remap = this.SVGremapMinus;
+        for (var i = 0, m = this.data.length; i < m; i++) {
+          if (this.data[i]) {
+            var child = svg.Add(this.data[i].toSVG(variant,svg.scale,remap),svg.w,0,true);
+            if (child.skew) {svg.skew = child.skew}
+            remap = null;
+          }
+        }
+        svg.Clean(); var text = this.data.join("");
+        if (svg.skew && text.length !== 1) {delete svg.skew}
+        if (svg.r > svg.w && text.length === 1 && !variant.noIC)
+          {svg.ic = svg.r - svg.w; svg.w = svg.r}
+	this.SVGhandleColor(svg);
+        this.SVGsaveData(svg);
+	return svg;
+      },
+    }),
 
     MML.mtext.Augment({
       toSVG: function () {
--- a/unpacked/extensions/MathML/content-mathml.js
+++ b/unpacked/extensions/MathML/content-mathml.js
@@ -190,7 +190,18 @@ MathJax.Extension["MathML/content-mathml"] = (function(HUB) {
     */
     appendToken: function(parentNode,name,textContent) {
       var element = CToP.createElement(name);
-      element.appendChild(document.createTextNode(textContent));
+      if (name === 'mn' && textContent.substr(0,1) === "-") {
+        //
+        // use <mrow><mo>&#x2212;</mo><mn>n</mn></mrow> instead of <mn>-n</mn>
+        //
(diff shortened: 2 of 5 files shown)


pkra:
>  Resolve issue #989.

Since this came from the Content MathML extension's output, should we fix it there as well (or am I missing the fix from some earlier patch)?

> This also tweaks the NativeMML output; should we do that? 

I'm ok with that. 

> Or make a configuration parameter to prevent that?

I don't mind either way. Perhaps it's more appropriate to be in v3.0 to have an explicit "tweak MathML" option (maybe a light-weight semantic enrichment module).

dpvc:
>  should we fix it there as well

Probably.  It doesn't look like anything was done about that.  I'll look into it further.

> Perhaps it's more appropriate to be in v3.0 to have an explicit "tweak MathML" option

Yes, that sounds good.

> maybe a light-weight semantic enrichment module

I'm not sure how light-weight it is.

The current arrangement does not change the internal MathML, only the MathML entered into the page.  Thus the "Show Math As..." menu shows the natural MathML, not the MathML tweaked for the particular browser (e.g., the hack we have for FF to handle mlabeledtr for example).  But I guess this is a different type of thing, where the input MathML can be improved.

dpvc:
> should we fix it there as well

OK, I added code to convert `<mn>-n</mn>` to `<mrow><mo>&#x2212;</mo><mn>n</mn></mrow>` in the content-mathml extension.

unpacked/jax/output/SVG/jax.js
```diff
@@ -1566,6 +1566,31 @@
 	return svg;
       }
     });
+    
+    MML.mn.Augment({
+      SVGremapMinus: function (text) {return text.replace(/^-/,"\u2212")},
+      toSVG: function () {
+        this.SVGgetStyles();
```
zorkow: Indentation?

zorkow:
I'd suggest a single utility function for the replacement of minus.
Also native/mml/svg is effectively 3 times the same code. Would it be worth the effort to combine those into a common utility method? Would make it easier to maintain.

unpacked/jax/output/SVG/jax.js
dpvc: I'm not sure I see the indentation problem here.

unpacked/jax/output/SVG/jax.js
```diff
@@ -1566,6 +1566,31 @@
...
+            if (child.skew) {svg.skew = child.skew}
+            remap = null;
+          }
+        }
+        svg.Clean(); var text = this.data.join("");
+        if (svg.skew && text.length !== 1) {delete svg.skew}
+        if (svg.r > svg.w && text.length === 1 && !variant.noIC)
+          {svg.ic = svg.r - svg.w; svg.w = svg.r}
+	this.SVGhandleColor(svg);
+        this.SVGsaveData(svg);
```
dpvc: Some lines have spaces for indenting and some have tabs (1589 has a tab and the surrounding ones have spaces).  The initial "+" here is throwing off the indenting, but in the actual file, it looks OK.  I prefer to do indenting with spacing but she of the files have tabs in them.  Do you think I should untabify the files?

dpvc:
Do you mean the `function (text) {return text.replace(/^-/,"\u2212")},`?  I'm not particularly worried about repeating this small function.  There turns out to be a lot of overlap among the three (or sometimes four) main output jax, and it is true that that duplication could be refactored.  It grew out of the fact that the initial two output jax were HTML-CSS and NativeMML, and they had very little overlap.  The addition of SVG made some of the duplication apparent, and now CommonHTML has made more of that.  I think it is probably not cost-effective to refactor for 2.x, but should do so for 3.0.

unpacked/jax/output/SVG/jax.js
zorkow: I see. Generally I like to avoid tabbing whitespace for exactly this reason.
But I don't think it makes sense to start doing this now in a single PR. Maybe for all files in a separate PR.

(setq-default indent-tabs-mode nil)
is you friend....

zorkow:
OK, I am fine with that.
LGTM.

unpacked/jax/output/SVG/jax.js
dpvc: I use that in emacs, but I don't always use emacs.  The tabs are coming from a different editor, in this case.  Not sure if I can configure it to use spaces or not.

Issue #1421, opened 2016-03-23
SVG data in fontdata-extra.js is incorrect


dpvc:
The widths of the single-character versions should be times 1000, but are given as decimals instead.  This is easily fixed, but needs to be done.

See mathjax/MathJax-node#202 for the original report.

pkra:
I had already filed https://github.com/mathjax/MathJax/issues/1418

pkra:
Oh nevermind. This is a new bug. Sorry!

dpvc:
The `font-fixes` branch of my fork of MathJax fixes this problem.

dpvc:
==> Merged

dpvc:
==> In testsuite

LaTeXToMathML/letters/issue1421.html

Issue #1418, opened 2016-03-21
SVG font data for `&#x2206;`


pkra:
From https://github.com/mathjax/MathJax-node/issues/196

> It looks like the SVG font data for `&#x2206;` is incorrect.

dpvc:
The `font-fixes` branch of my fork of MathJax includes the fix for this.

dpvc:
==> Merged

dpvc:
==> In testsuite

LaTeXToMathML/letters/issue1418.html

Issue #1187, opened 2015-05-08
positioning of subscript or superscript minus


orangeplanet:
A subscript or superscript minus sometimes (but not always) wanders too far from the target.  It seems to especially happen when the neighboring superscript/subscript slot is occupied at the same time. Adding curly braces to whatever is on text line seems to negate the effect, but I'm not sure why.  

See https://jsfiddle.net/s830g49p/

dpvc:
The issue is a little complicated.  TeX's spacing rules sometimes rely on the font metrics being set up in a particular way, and this turns out to be one of them.  In the TeX fonts, the height and depth of the minus are set to be the same as that of a plus (even though the minus doesn't take up that amount of vertical space), and that has an impact in the case of a base with both super and subscript (where the spacing between the two is based on the height and depth of the subscript and superscript).  MathJax's fonts don't have the extra space around the minus sign (the bounding boxes are tight for all the characters).

It is possible to adjust the bounding box information used by MathJax, but in this case, you have to be a bit careful, as the minus sign is used for several purposes (like extending arrows, and making horizontal lines for things like square roots, `\overline`, and `\underline`.  So while changing the bounding box information for the minus would fix the super- and sub-script problem, it would cause other things to be spaced improperly.

It is easy to change the character used for the horizontal line to a different one (like U+00AF), so that is not really a problem.  But changing the height and depth of the minus will affect the height and depth of stretched arrows, and that is harder to fix.

Here is a work-around you can use for now, if you don't mind the slight change in positioning of stretched arrows.  Add the following to your page just *before* the script that loads MathJax.js itself.

``` html
<script>
MathJax = {
  AuthorInit: function() {
    MathJax.Hub.Register.LoadHook("[MathJax]/jax/output/HTML-CSS/fonts/TeX/fontdata.js",function () {
      var FONTDATA = MathJax.OutputJax["HTML-CSS"].FONTDATA;
      var MAIN = FONTDATA.FONTS.MathJax_Main;
      MAIN[0x2212][0] = MAIN[0x002B][0];
      MAIN[0x2212][1] = MAIN[0x002B][1];
      FONTDATA.RULECHAR = 0x00AF;
      FONTDATA.DELIMITERS[0x2212] = {alias: 0x00AF, dir: "H"};
    });
    MathJax.Hub.Register.LoadHook("[MathJax]/jax/output/HTML-CSS/fonts/STIX/fontdata.js",function () {
[...]

Issue #1175, opened 2015-04-28
\lvert and \rvert should be to U+007C and \lVert and \rVert to U+2016


dpvc:
Currently these map to U+2223 and U+2225, which are not appropriate for delimiters.

dpvc:
The `font-fixes` branch of my fork of MathJax includes the change.

dpvc:
==> Merged

dpvc:
==> In testsuite

LaTeXToMathML/delimiters/issue1175.html

PR #1529, opened 2016-06-17
Font fixes

dpvc:
This pull request includes fixes for a number of font issues in the 2.x milestone.

* Make U+222B (integral) stretchy in the STIX fonts (#1092)
* Remap `|` to variant form in STIX (that has descender) and map variant to original form (#1154)
* Use U+007C and U+2016 for delimiters rather than U+2223 and U+2225 (#1175)
* Fix SVG font data for stretchy characters in the TeX font data-extras files  (#1421)
* Alias U+2206 to U+0394 and remove incorrect U+2206 from SVG font files (#1418)
* Make height and depth of minus match that of plus (needed for TeX super/subscript algorithm to work properly), and adjust for that when it is used as an extender in stretchy characters (#1187).

The last one affected all the web font files, in the HTML-CSS, SVG, and CommonHTML output, so it was extensive.  This doesn't solve all the problems in the extra web font data files (there are many issues).  I also didn't try to fix the scripts that were used originally to create the files, as I didn't want to waste time on that, as I think the whole web font approach needs to be rethought for v3.0.  So these were edited by hand (we'll need to remember that for the future).


zorkow:
I did not really understand what is going on... but I assume it works.

dpvc:
Yes, I think it does.  If not, I'm hoping the test suite will catch it.  So I will merge for now.

PR #1530, opened 2016-06-20
Fix problem with Safari inserting line breaks before in-line math. #1478

dpvc:
Fix problem with Safari inserting line breaks before in-line math due to new width-detection scheme.  The `display:table-cell` that was used caused Safari to leave a line break after the table-cell is removed.  The solution is to enclose it in an element with `display:table`.  This still allows the detection of the correct with compared to floating material, while still being able to work inside shrink-wrap element.  #1478.


zorkow:
lgtm.

PR #1534, opened 2016-06-22
Make sure CHTML output stays a table-cell when focused.

dpvc:
This fixes a problem in Firefox with CHTML output that includes labeled rows.  When they gain focus, they could change size (due to the changes in commit 97ab12e).  This prevents that from happening. Other browsers don't seem to have this problem.


zorkow:
lgtm.

PR #1537, opened 2016-06-24
Make mhchem have a config parameter to select the new version

dpvc:
Make mhchem have a config parameter that can be used to select the new version (when it is available on the CDN) or the old legacy version.  The default is the legacy version, so this is an opt-in for the new mhchem, which is what I think we agreed on.  The parameter is named `legacy`, but I am open to other suggestions.

I had to put the old version inside an `if-then` statement, but I didn't re-indent it, as I didn't want the difference to look extensive than it is.  Volker may want me to change that.  :-)

### DIFF
--- a/unpacked/extensions/TeX/mhchem.js
+++ b/unpacked/extensions/TeX/mhchem.js
@@ -26,8 +26,18 @@
  */
 
 MathJax.Extension["TeX/mhchem"] = {
-  version: "2.6.0"
+  version: "2.6.0",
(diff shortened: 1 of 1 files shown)


pkra:
> The parameter is named `legacy`, but I am open to other suggestions.

OK with me but I also don't have strong feelings about it.

zorkow:
I think this is for @pkra as he is been in the mhchem loop.

pkra:
Sorry I missed my mention.

LGTM

Issue #1364, opened 2016-01-30
Improper rendering with nested roots


zopyx:
See test #13

https://print-css.rocks/_static/lesson-mathml-torture-test/vivliostyle-output.pdf

Source is here

https://github.com/zopyx/css-paged-media-tutorial/tree/master/lesson-mathml-torture-test

dpvc:
I can not reproduce your result.  The output is fine for me in all output modes for my own test page.  The example page you point to doesn't include the MathJax configuration that Vivliostyle uses, so I don't know how MathJax is configured.  Do you have automatic line breaks turned on?  It looks like your result may be do to line-breaing within the roots that would have caused the table to be too wide.

pkra:
Thanks for following up downstream, @kwkbtr. Let us know if we can help with anything.

kwkbtr:
Currently Vivliostyle always enables the automatic line breaks.
The configuration is here:
https://github.com/vivliostyle/vivliostyle.js/blob/2016.1/resources/mathjax-config.js#L12

I confirmed that the math is rendered correctly when the automatic line breaks is turned off.

I filed an issue to Vivliostyle (vivliostyle/vivliostyle.js#167), but I'm not sure if this problem is specific to Vivliostyle. Even without Vivliostyle, the same problem occurs when width of the table column *before* typesetting is narrower than width required for the typeset math to fit inside the column.

kwkbtr:
> Even without Vivliostyle, the same problem occurs when width of the table column before typesetting is narrower than width required for the typeset math to fit inside the column.

You can see this by removing table rows except the row "13" from the sample (and turning on the automatic line breaks).

pkra:
Thanks @kwkbtr Is there any chance somebody could create a minimal sample, e.g., on codepen, jsbin?

kwkbtr:
> Is there any chance somebody could create a minimal sample, e.g., on codepen, jsbin?

Created one: http://jsbin.com/vojimahuno/edit?html,output

pkra:
Great! Thanks, @kwkbtr.

dpvc:
There are several issues going on, here.  The first is that line breaking can occur inside a square root, and that is what is happening here, because MathJax things the line width is small (more on this below).  The result is pretty ugly, but there isn't much voice about how this equation can be broken.  Perhaps it would help to left-jutify rather than center the parts, or indent the second line, but I'm note sure.  In any case, that part is "expected behavior" for line breaking this expression.

[...]

PR #1538, opened 2016-06-24
Don't let preview width interfere with the determination of the container width. #1364

dpvc:
Don't let preview width interfere with the determination of the container width in HTML-CSS, CHTML, and SVG output.  Resolves issue #1364.


zorkow:
lgtm.

Issue #1541, opened 2016-07-04
[CommonHTML] regression on develop branch: horizontal stretch causes processing error


pkra:
I ran into a regression on the develop branch.

Here's a minimal example. It seems to require the mtable and the mover (this came from an AMScd source).

```html
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
  <mtable rowspacing="5pt" columnspacing="5pt" displaystyle="true">
    <mtr>
      <mtd>
        <mi>F</mi>
      </mtd>
      <mtd>
        <mover>
          <mo minsize="2.75em">&#x2192;</mo>
        </mover>
      </mtd>
      <mtd> 
        <mi>F</mi>
      </mtd>
    </mtr>
  </mtable>
</math>
```

`lastError` gave me:

```console
TypeError: Cannot read property 'firstChild' of null
    at CHTMLstretchH (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/jax.js?rev=2.6.1:1959:22)
    at CHTMLstretchCoreH (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/jax.js?rev=2.6.1:1472:28)
    at CHTMLstretchH (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/jax.js?rev=2.6.1:1704:14)
    at CHTMLstretchCoreH (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/jax.js?rev=2.6.1:1472:28)
    at CHTMLstretchH (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/jax.js?rev=2.6.1:1468:36)
    at CHTMLstretchCells (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/autoload/mtable.js?rev=2.6.1:422:55)
    at toCommonHTML (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/autoload/mtable.js?rev=2.6.1:71:12)
    at CHTMLaddChild (https://rawgit.com/mathjax/MathJax/develop/unpacked/jax/output/CommonHTML/jax.js?rev=2.6.1:1395:25)
[...]

PR #1542, opened 2016-07-04
Fix problems with stretching <mover> in <mtd> elements.  #1541

dpvc:
Fix `getNode()` yet again to walk the tree properly.  Also, make `CHTMLcoreNode()` step through inferred rows (to match what `Core()` does) so that it will get the correct DOM node for the `Core()` element.  Finally, make `CHTMLaddChild()` add a box of the correct type if `forceChild` is set.  Resolves issue #1541.


zorkow:
lgtm.

Issue #1548, opened 2016-07-04
Square roots containing line breaks don't render properly


dpvc:
The example

```
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
  <msqrt>
    <mtext mathvariant="italic">aaaaa</mtext>
    <mspace linebreak="newline"></mspace>
    <mtext mathvariant="italic">b</mtext>
  </msqrt>
</math>
```

doesn't render properly in any of the HTML-CSS, CHTML, or SVG output formats.  (Each fails in a different way, but all three have the two lines centered, when they should be left-justified.)

dpvc:
This is fixed in the `issue1548` branch of my fork of MathJax.

dpvc:
==> Merged.

dpvc:
==> In testsuite

MathMLToDisplay/Topics/LineBreak/issue1548.html

PR #1549, opened 2016-07-04
Fix determination of line-breaing parent element. #1548.

dpvc:
Fix determination of line-breaking parent element.  It was moving too high in some cases.  E.g., in

```
  <mrow>
    <msqrt>
      ...
    </msqrt>
  </mrow>
```

the inferred mrow in the msqrt was stepping out to the msqrt (correctly) but then stepped out to the mrow and its parent (since it was an mrow with one child).  The change to `parent.isEmbellished()` causes the stepping out to occur only when the outer mrow represents an embellished operator.  (We only want `isTop` to be true in that case.)

Resolves issue #1548.


zorkow:
lgtm.

Issue #1543, opened 2016-07-04
[CommonHTML] regression: vector arrow not properly centered


dpvc:
On the develop branch, `\vec x` is not centered in CommonHTML output.

dpvc:
The `issue1543` branch of my fork of MathJax resolves this.

dpvc:
==> Merged

PR #1550, opened 2016-07-05
Fix vector arrow positioning in CHTML.  #1543

dpvc:
Change class to mjx-box so that width will have the proper effect (due to fix in CSS in PR #1448).  Resolves issue #1543.

### DIFF
--- a/unpacked/jax/output/CommonHTML/jax.js
+++ b/unpacked/jax/output/CommonHTML/jax.js
@@ -1882,7 +1882,7 @@
         //  something, so put them over a space and remove the space's width
         //
         node = node.firstChild;
-        var space = CHTML.Element("mjx-span",{style:{width:".25em","margin-left":"-.25em"}});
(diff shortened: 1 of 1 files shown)


unpacked/jax/output/CommonHTML/jax.js
```diff
@@ -1882,7 +1882,7 @@
         //  something, so put them over a space and remove the space's width
         //
         node = node.firstChild;
-        var space = CHTML.Element("mjx-span",{style:{width:".25em","margin-left":"-.25em"}});
+        var space = CHTML.Element("mjx-box",{style:{width:".25em","margin-left":"-.25em"}});
```
zorkow: I assume this does not have any implications on CSS behaviour?

dpvc: Only the desired one of making it work. :-) 

The only CSS that these two set is the `display` value, and that is what needs to be changed.  So there are no side-effects of this.

zorkow:
lgtm.

Issue #1087, opened 2015-03-10
<wbr> tags


mark-wade:
I am a developer for a forum software which, to prevent long user-provideds topic titles with no spaces from breaking the layout, inserts wbr tags every 20 characters in words in titles that are longer than that. For example, if a user creates a title "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", it will get rendered as "aaaaaaaaaaaaaaaaaaaa-wbr-aaaaaaaaaaaaaaaaaaaa" (am using "-wbr-" rather than the actual tag as GitHub seems to parse it)

A bug report was submitted to us that this causes MathJax to not work correctly: http://community.invisionpower.com/4bugtrack/wbrwbr-randomly-added-to-topic-titles-after-upgrading-r3931/

The HTML5 <wbr> tag represents a word-break oppertunity within text. Since MathJax modifies the content, it would be appropriate for it to ignore such tags within the content it intends to replace.

pkra:
Unfortunately, TeX-like syntax and HTML markup do not play well together.

For sanity reasons, the tex2jax preprocessor stops parsing at HTML elements. I don't think it's likely we will change this behavior anytime soon (if ever).

If the situation is restricted enough, I suppose you could do some regexp clean up before `tex2jax.js` starts up but then again it might be cleaner to isolate the TeX while saving (since the forum software modifies the content anyway).

mark-wade:
Could you direct me to where abouts in the code that's done?

pkra:
Which part are you referring to? The tex2jax part or the "do it yourself" part? For the latter, just run whatever works for your setup before MathJax.js is loaded, e.g., as part of a custom configuration (see http://docs.mathjax.org/en/latest/configuration.html#using-in-line-configuration-options).

mark-wade:
Wherever it halts on a HTML tag.
As an aside, someone mentioned to me that it ignores br tags... is that right? If so, would it actually be terribly dramatic to alter that to also do whatever it does there for wbr?

pkra:
Hm... did I put my foot in my mouth? [Looks](http://docs.mathjax.org/en/latest/tex.html#tex-and-latex-in-html-documents) like [it](https://github.com/mathjax/MathJax/blob/master/unpacked/extensions/tex2jax.js#L167). So now we're in a slippery-slope discussion. @dpvc ?

dpvc:
[...]

PR #1551, opened 2016-07-06
Allow <wbr> in TeX code (just like <br>).  #1087

dpvc:
Add `<wbr>` as an ignored tag, and change to a list of ones to ignore.  Cache the `math.nextSibling` node.  Remove `msieNewlineBug` and fold that into the `ignoreTags` hash.  Resolves issue #1087.


zorkow:
lgtm

Issue #1022, opened 2015-01-27
linebreaks incompatible with href using latex input and HTML-CSS output


lionelperrin:
When I turn on linebreaks: { automatic: true } then the links in my formula are not rendered.
When I turn off linebreaks, the links are rendered as expected.

Here follows an example of tex formula:
$$ \href(/audit/45, A) + \href(/audit/46, B) + \href(/audit/47, C) $$

I've reproduced with MathJax 2.3.0 and MathJax 2.4.0

Regards,

Lionel

pkra:
I'm unable to reproduce this. Can you share a live sample (e.g., on jsbin, codepen etc) where this fails?

FWIW, the syntax is ` $$ \href{url}{math} $$` ([see the docs](http://docs.mathjax.org/en/latest/tex.html#html))

lionelperrin:
You can try with the following code. I reproduced it on jsbin: http://jsbin.com/lizanadiro/3/

In case you need it:
```html
<head><script type="text/x-mathjax-config">
MathJax.Hub.Config({
  "HTML-CSS": {
    linebreaks: { automatic: true },
  }
  });
  </script>
  <script type="text/javascript"
  src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
</script>
</head>
<body>
  <div style='width: 500px'>
  $$ \href{http://docs.mathjax.org/}{SomethingQuiteLongTooFillTheWholeScreen} + \href{http://docs.mathjax.org/}{SomethingQuiteLongTooFillTheWholeScreenToo} + \href{http://docs.mathjax.org/}{AnotherSomethingQuiteLongTooFillTheWholeScreen} + \href{http://docs.mathjax.org/}{And A LastSomethingQuiteLongTooFillTheWholeScreen} $$
  </div>
</body>
```

pkra:
Thanks for following up. Strange; I could've sworn I did something similar yesterday and failed to reproduce. Oh well, I can reproduce it in your example. 

Thanks so much for the report. We'll look into it.

dpvc:
The `issue1022` branch of my fork of MathJax handles the hrefs during line breaks.

dpvc:
==> Merged.

dpvc:
==> In testsuite

MathMLToDisplay/Topics/LineBreak/issue1022.html

PR #1552, opened 2016-07-06
Handle hrefs during line breaking.  #1022

dpvc:
Fix line breaking to handle hrefs properly both when they are split and when they are not.  This affects HTML-CSS, CHTML, and SVG output.  Resolves issue #1022.


zorkow:
lgtm

Issue #1183, opened 2015-05-03
Safe extension should process the padding value of \bbox


dpvc:
Currently, `\bbox[10px]{math}` doesn't pass the `10px` through any filter, but it should (especially if #1182 is implemented).

dpvc:
The `issue1182` branch of my fork of MathJax adds support for this.

dpvc:
==> Merged

dpvc:
==> In testsuite

Configuration/Safe/issue1183.html

Issue #1182, opened 2015-05-03
Safe extension should allow max/min values for numeric properties like padding


dpvc:
This would allow some use of the properties, but not excessive use.  This was requested on the MathJax-dev list.

dpvc:
The `issue1182` branch of my fork of MathJax adds this feature.

dpvc:
==> Merged

dpvc:
==> In testsuite

Configuration/Safe/issue1182-1.html
Configuration/Safe/issue1182-2.html

PR #1554, opened 2016-07-07
Add filtering of CSS styles like padding/margin, etc.  #1182 and #1183

dpvc:
* Add filtering for CSS styles `padding`, `margin`, `border`, `outline`, etc. (#1182)
* Add handling of `padding-top`, `padding-right`, etc. based on `safeStyles` settings for `padding` as well as for `paddingTop`, etc.
* Add max/min values for `font-size` style.
* Add filtering for padding in `\bbox` (#1183)

### DIFF
--- a/unpacked/extensions/Safe.js
+++ b/unpacked/extensions/Safe.js
@@ -395,7 +395,11 @@
     //  Filter the styles for \bbox
     //
     TEX.Parse.Augment({
-      BBoxStyle: function (styles) {return SAFE.filterStyles(styles)}
+      BBoxStyle: function (styles) {return SAFE.filterStyles(styles)},
+      BBoxPadding: function (pad) {
+        var styles = SAFE.filterStyles("padding: "+pad);
+        return (styles ? pad : 0);
+      }
     });
 
   });
--- a/unpacked/extensions/Safe.js
+++ b/unpacked/extensions/Safe.js
@@ -43,6 +43,7 @@
     },
     sizeMin: .7,        // \scriptsize
     sizeMax: 1.44,      // \large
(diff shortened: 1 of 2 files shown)


unpacked/extensions/Safe.js
```diff
@@ -319,7 +395,11 @@
     //  Filter the styles for \bbox
     //
     TEX.Parse.Augment({
-      BBoxStyle: function (styles) {return SAFE.filterStyles(styles)}
+      BBoxStyle: function (styles) {return SAFE.filterStyles(styles)},
+      BBoxPadding: function (pad) {
+        var styles = SAFE.filterStyles("padding: "+pad);
+        return (styles ? pad : 0);
```
zorkow: Remove obsolete parens.

dpvc: Personally, I prefer these parentheses, and there are a number of places that they are used in this way in the current code base.  We can <s>argue about</s> develop a coding style for v3.0, but I don't think is necessary to switch styles for a few lines of 2.7 when all the other instances are different.

zorkow: Fine. NP.

Personally I prefer redundant whitespace to obsolete parentheses for readability....
I believe for v3.0 we just have to agree on a style and mandatory linting!

pkra: >  I believe for v3.0 we just have to agree on a style and mandatory linting!

And use editorconfig.org and similar tools to minimize inconvenience.

dpvc: > Personally I prefer redundant whitespace

I personally find excessive white space hard to read.  But I know I will have to modify my old stodgy ways.  I will try not to argue too much, and am sure I can adapt over time.

Issue #1434, opened 2016-04-10
[Menu] move from offsetWidth to getBoundingClientRect for positioning [was `.MathJax_Menu` does not exceed a `left` property of >627]


DizzyZane:
Test case: http://dizzyzane.web.fc2.com/markdown/tests/tests.html & http://dizzyzane.web.fc2.com/markdown/tests/tests.html#top (same page, different colouration)

The menu will not go into the right half of the column on a 1280px wide monitor.

Chrome version: Version 51.0.2704.0 canary (64-bit)

Firefox: 45.0.1

dpvc:
MathJax tries to keep the menu within the space allocated to the `document.body` so that the menu is shown within the space of the document when possible (e.g., if the document is in an iframe, it needs to stay within that iframe's space).

Your `document.body` has CSS that sets `max-width: 768px`, so when MathJax asks for the `document.body.offsetWidth`, it gets 768 and thinks that is the size of the space it has to work with.  unfortunately, with an `auto` margin, the `document.body.offsetLeft` is still 0, not the size of the margin, and so MathJax thinks that the document body is flush left.

It may be possible to use `getBoundingClientRect()` to get the correct offset of the body.  The computation of the correct position is pretty fragile, and there are a number of situations like this where it fails.  It may be possible to resolve it in a future release.

@Peter, this might also help with the incorrect positioning of the menu in Readium.

pkra:
@dpvc Readium turned off the menu I think. But it might be useful for the next iteration of the menu.

dpvc:
The `issue1434` branch of my fork of MathJax includes code to better handle the situation when the document body has margins or is scrolled.

dpvc:
==> Merged

dpvc:
==> In testsuite

UI/math-menu/issue1434.html

PR #1556, opened 2016-07-12
Improve menu placement.  #1434

dpvc:
Use `getBoundingClientRect()` to improve handling of menu placement when the body is scrolled or has margin settings.  Resolves issue #1434.

Here `bbox.right` gives the position of the right-hand side of the body, unless it is scrolled left, in which case the `bbox.left` will be negative, so we add it in.  These don't seem to take the left margin into account, so we add that in as well.

### DIFF
--- a/unpacked/extensions/MathMenu.js
+++ b/unpacked/extensions/MathMenu.js
@@ -313,6 +313,9 @@
       this.posted = true;
       if (menu.offsetWidth) menu.style.width = (menu.offsetWidth+2) + "px";
       var x = event.pageX, y = event.pageY;
+      var bbox = document.body.getBoundingClientRect();
+      var styles = window.getComputedStyle(document.body);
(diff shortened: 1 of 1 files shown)


unpacked/extensions/MathMenu.js
```diff
@@ -313,6 +313,9 @@
       this.posted = true;
       if (menu.offsetWidth) menu.style.width = (menu.offsetWidth+2) + "px";
       var x = event.pageX, y = event.pageY;
+      var bbox = document.body.getBoundingClientRect();
+      var styles = (window.getComputedStyle ? window.getComputedStyle(document.body) : {marginLeft: "0px"});
```
zorkow: Remove obsolete parens.

dpvc: See [this comment](https://github.com/mathjax/MathJax/pull/1554/files/081ff8dfac471731aecb7656fbea407f1a6f8986#r71977474).

zorkow:
lgtm

Issue #911, opened 2014-09-11
Spacing following \tfrac in superscript is wrong


dpvc:
The spacing between the fraction and plus is too big for

    e^{\tfrac{1}{2} + c}

This is due to the check for `prevLevel > 0` in the `texSpacing` method.  Probably checking the operators `scriptlevel` is sufficient (or using an "or" rather than "and" for the check).

pkra:
Just adding this to the milestone so that it's on some milestone. Feel free to move it to another milestone.

dpvc:
The `issue911` branch of my fork of MathJax resolves this.

dpvc:
==> Merged

dpvc:
==> In testsuite

MathMLToDisplay/Presentation/ScriptsAndLimits/issue911.html

PR #1567, opened 2016-07-23
Fix spacing for initial fraction in exponent position.  #911

dpvc:
Don't require both `prevLevel > 0` and `scriptlevel > 0` (either true should be sufficient) when determining TeX spacing.  This gets the spacing right for `e^{\tfrac{1}{2} + c}`.

Resolves issue #911.


dpvc:
Sorry, I forgot this one (only noticed it when the pull requests were few enough to see what branches hadn't been merged).

zorkow:
lgtm.

Issue #1568, opened 2016-07-27
Parsing problems with \def with templates


dpvc:
Currently, a template can match in the middle of a control sequence name.  For example

```
\def\x#1b#2.{\texttt{#1}}
\x\abc.
```
will output `\a` since the `b` matches in the middle of the `\abc`.

Also, the code
```
\def\x#1\a#2.{\texttt{[#1][#2]}}
\x abc \abc abc \a def .
```

will output `[abc ][bc abc \a def]` rather than the expected `[abc \abc abc ][ def ]`.

dpvc:
Both issues are resolved in the `issue1568` branch of my fork of MathJax.

dpvc:
==> Merged

dpvc:
==> In testsuite

LaTeXToMathML/macro/issue1569.html

PR #1569, opened 2016-07-28
Treat control sequences as a unit when matching a macro template. #1568

dpvc:
Treat control sequences as a unit when matching a macro template.  This applies both to skipping over control sequences that *don't* match, and making sure a *matched* sequence is a complete token.

Resolves both issues for #1568.


zorkow:
lgtm

Issue #1573, opened 2016-08-04
[SVG, HTML-CSS output] menclose does not inherit color


pkra:
Via the user group. 

For SVG, it looks like we forget `menclose` in https://github.com/mathjax/MathJax/issues/1002. This seems to be it: https://github.com/mathjax/MathJax/blob/master/unpacked/jax/output/SVG/autoload/menclose.js#L116.

For HTML-CSS it seems to be https://github.com/mathjax/MathJax/blob/master/unpacked/jax/output/HTML-CSS/autoload/menclose.js#L58

pkra:
Edited my OP with information for HTML-CSS.

dpvc:
The `issue1573` branch of my fork of MathJax includes the fix.  Thanks for finding the lines that caused the problem.

dpvc:
==> Merged

dpvc:
==> In testsuite

MathMLToDisplay/Presentation/GeneralLayout/menclose/issue1573.html

PR #1574, opened 2016-08-04
Use currentColor rather than black for menclose with no math color.  #1573

dpvc:
Use currentColor rather than black for menclose when mathcolor isn't specified.

Resolves issue #1573.

### DIFF
--- a/unpacked/jax/output/SVG/autoload/menclose.js
+++ b/unpacked/jax/output/SVG/autoload/menclose.js
@@ -113,7 +113,6 @@ MathJax.Hub.Register.StartupHook("SVG Jax Ready",function () {
       t = Math.max(1/SVG.em,t);  // see issue #414
       var H = base.h+p+t, D = base.d+p+t, W = base.w+2*(p+t);
       var dx = 0, w, h, i, m, borders = [false,false,false,false];
-      if (!values.mathcolor) {values.mathcolor = "currenColor"}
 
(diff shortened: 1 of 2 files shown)


pkra:
LGTM.

unpacked/jax/output/SVG/autoload/menclose.js
```diff
@@ -113,7 +113,7 @@ MathJax.Hub.Register.StartupHook("SVG Jax Ready",function () {
       t = Math.max(1/SVG.em,t);  // see issue #414
       var H = base.h+p+t, D = base.d+p+t, W = base.w+2*(p+t);
       var dx = 0, w, h, i, m, borders = [false,false,false,false];
-      if (!values.mathcolor) {values.mathcolor = "black"}
+      if (!values.mathcolor) {values.mathcolor = "currenColor"}
```
pkra: ( :frowning: Sorry I missed that.)

dpvc: I missed it, too.  The code worked because when the color attribute was set, it didn't match a valid color and was ignored, meaning that the color was inherited anyway.  I'll fix it.  We might be able to leave that line out entirely.

dpvc:
It turns out we don't need the `currentColor` line in the SVG output after all, but we do in the HTML-CSS output.

I don't think this needs more review, so I'm merging.

Issue #1586, opened 2016-08-17
[TeX input] \operatorname {Res\ } throws error


pkra:
From https://phabricator.wikimedia.org/T136812. 

In a quick test, this didn't cause an error in real LaTeX (but I couldn't check warnings). 

Is this expected behavior?

dpvc:
It's a bug (related to #1563).  Arguments to some macros have leading and trailing spaces removed, and the trimming meant that the space from `\ ` was removed, leaving it essentially `{Res\}`, and so the closing brace was escaped, and that messed up the brace balancing.

The `issue1586` branch of my fork of MathJax includes a fix.

dpvc:
==> In testsuite

LaTeXToMathML/issue1586.html

physikerwelt:
@dpvc Thank you. We cherry picked your change and it now works for wmf cf.
https://en.wikipedia.org/api/rest_v1/media/math/render/mml/a2c9f30d6ff4230bf5406de1fc4ccdb6ba2712b0

dpvc:
@physikerwelt, glad you were able to get it working for you.

dpvc:
==> Merged

Issue #1563, opened 2016-07-20
\def and \let don't work with \  (macro whose name is a space)


dpvc:
The command `\def\ {x}` and `\let\ =x` are processed, but don't change the macro `\ ` (they actually set a macro with no name).  This is due to timing whitespace in `GetCSname()`.

dpvc:
This is fixed in the `issue1563` branch of my fork of MathJax.

dpvc:
==> Merged.

dpvc:
==> In testsuite

LaTeXToMathML/macro/issue1563.html

PR #1587, opened 2016-08-17
Make sure trimSpaces() doesn't remove tailing space in '\ '.  #1586

dpvc:
Make sure trimSpaces() doesn't remove tailing space in '\ '.  Resolves issue #1586 (and handles #1563 better).


zorkow:
lgtm.

Issue #1555, opened 2016-07-08
MathJax_Preview causes extra line break in Firefox within list items


pkra:
From a user report elsewhere.

> If one starts a list item `<li>` with a mathjax alignment structure, the equations will become vertically centered with respect to the list item in Chrome and top aligned in Firefox.
> You can see this in https://cdi2tp.math.tecnico.ulisboa.pt/enunciados/ExameAsol2015160627.php if you open the second collapsed `<details>`.

Firefox:

![Firefox](https://cloud.githubusercontent.com/assets/426686/16685775/abc90686-450d-11e6-9c95-ec5a1c45a3dd.png)

Chrome 

![Chrome](https://cloud.githubusercontent.com/assets/426686/16685779/b0a9546c-450d-11e6-99a8-3f657807213d.png)

I haven't had a chance to take a closer look, so labeling it `investigate`.

cebola2:
I have created a more minimal example that exhibits the same behavior. There are no <details></details> in this case hence the title for this issue seems to be a misnomer.
See https://math.tecnico.ulisboa.pt/teste

dpvc:
The issue is that Firefox and Chrome treat the empty MathJax_Preview span that precedes the typeset math differently in this situation.  In Firefox, the empty span is treated as in-line material and so the block level expression needs to start a new line following that material.  In Chrome, the fact that the span is empty seems to be taken into account and Chrome puts the block-level material at the top of the list item.

If the MathJax_Preview span is removed, Firefox's output is the same as Chrome's.  So adding he following

```
<script type="text/x-mathjax-config">
Mathjax.Hub.Queue(function () {
  var previews = document.querySelectAll(".MathJax_Preview");
  for (var i = 0, m = previews.length; i < m; i++) {
    previews[i].parentNode.removeChild(previews[i]);
  }
});
</script>
```
to the page somewhere *before* the script that loads MathJax.js should clear up the issue.

@pkra, do we want to think about removing the previews ourselves?  There are reasons these weren't removed which we can discuss if you want to consider it.

pkra:
> @pkra, do we want to think about removing the previews ourselves? 

Yes, let's discuss this as the next F2F.

dpvc:
Actually, it turns out that setting `display:none` also has the desired effect, so that might be the better approach.

pkra:
:+1:

dpvc:
@pkra, do we want to add this to 2.7, since the change is minimal?

[...]

PR #1590, opened 2016-08-21
Set previews to have display:none.  #1555

dpvc:
Set previews to have `display:none` when their expressions have been typeset.  This avoids the problem with Firefox where the preview causes an extra line break when a displayed equation appears as the first thing in a list.

Resolves issue #1555.


zorkow:
lgtm.

Issue #1575, opened 2016-08-05
content-mathml extension throws an error for this expression


dpvc:
This expression

```
<math xmlns="http://www.w3.org/1998/Math/MathML">
  <apply>
    <plus/>
    <cn> 1 </cn>
    <apply>
      <times/>
      <cn> -444 </cn>
      <ci> x </ci>
    </apply>
  </apply>
</math>
```

causes the content-mathml to throw an error.  It looks like it is [line 1321](https://github.com/mathjax/MathJax/blob/develop/unpacked/extensions/MathML/content-mathml.js#L1321) which should be `children[1].textContent = -n;`, though the comment does say that this is modifying the original document, so it should probably be rewritten to avoid that.

Also, the `( n = Number(CToP.getTextContent(children[1])) < 0)` in the preceding `if` statement should be `( n = Number(CToP.getTextContent(children[1]))) < 0` (otherwise, `n` gets the value of  boolean rather than the number).

dpvc:
The `issue1575` branch of my fork of MathJax includes a fix for all three problems (the processing error, the modification of the document, and the bad value for `n`).

dpvc:
==> In testsuite

MathMLToDisplay/Content/issue1575.html

dpvc:
==> Merged.

PR #1592, opened 2016-08-21
Fix incorrect assignment in content-mathml extension.  #1575

dpvc:
Fix incorrect assignment in content-mathml extension, and don't modify original document.

Resolves issue #1575.

### DIFF
--- a/unpacked/extensions/MathML/content-mathml.js
+++ b/unpacked/extensions/MathML/content-mathml.js
@@ -1319,9 +1319,7 @@ MathJax.Extension["MathML/content-mathml"] = (function(HUB) {
               CToP.applyTransform(mrow,children[1],2);
             } else if (arg.nodeName === 'apply' && children.length>2 && children[0].nodeName === 'times' && children[1].nodeName === 'cn' && (n = Number(CToP.getTextContent(children[1]))) < 0) {
               CToP.appendToken(mrow,'mo','\u2212');
-              arg = arg.cloneNode(true);
-              children = CToP.getChildren(arg);
-              children[1].textContent = -n;
+              children[1].textContent = -n;     // OK to change MathML since it is being discarded afterward
               CToP.applyTransform(mrow,arg,2);
             } else{
               CToP.appendToken(mrow,'mo','+');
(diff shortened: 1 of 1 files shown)


unpacked/extensions/MathML/content-mathml.js
```diff
@@ -1317,9 +1317,11 @@ MathJax.Extension["MathML/content-mathml"] = (function(HUB) {
             } else if (arg.nodeName === 'apply' && children.length === 2 && children[0].nodeName === 'minus') {
               CToP.appendToken(mrow,'mo','\u2212');
               CToP.applyTransform(mrow,children[1],2);
-            } else if (arg.nodeName === 'apply' && children.length>2 && children[0].nodeName === 'times' && children[1].nodeName === 'cn' && ( n = Number(CToP.getTextContent(children[1])) < 0)) {
+            } else if (arg.nodeName === 'apply' && children.length>2 && children[0].nodeName === 'times' && children[1].nodeName === 'cn' && (n = Number(CToP.getTextContent(children[1]))) < 0) {
               CToP.appendToken(mrow,'mo','\u2212');
-              CToP.getTextContent(children[1]) = -n;// fix me: modifying document
+              arg = arg.cloneNode(true);
```
zorkow: Looks expensive, but I assume it is necessary to avoid changes to the orig.

zorkow:
o/w lgtm.

unpacked/extensions/MathML/content-mathml.js
dpvc: Yes, that was the reason for it.  I suppose one could change the document, do the transform, then change it back.  Perhaps it doesn't really matter,  since the MathML has been removed from the page anyway, and I think we are working on a copy that has been parsed from the stringified MathML.  Should I just change the document?

dpvc:
Your comment reminded me that the MathML that content-mathml is processing is actually a temporary MathML tree already, so there is no problem with modifying it, as it is discarded in the end (content-mathml actually replaces it with its result).  So I've removed the cloning and just changed the child directly.

Issue #1594, opened 2016-08-21
[CommonHTML] elements with font specified are not properly scaled


dpvc:
Specifying a font-family in CommonHTML causes the element to be scaled improperly.  E.g.

    \style{font-family:inherit}{x}
    \unicode[Arial]{xE0}
    \mmlToken{mi}[fontfamily="Arial"]{x}

all cause the elements to be scaled when compared to the math font.  This is not the case for HTML-CSS and SVG output.

dpvc:
The `issue1594` branch of my fork of MathJax includes a fix for this.

dpvc:
==> In testsuite

MathMLToDisplay/issue1594.html

dpvc:
==> Merged

PR #1595, opened 2016-08-21
Properly scale elements with font-family specified. #1594

dpvc:
Properly scale elements with font-family specified in CommonHTML.  Resolves issue #1594.


zorkow:
lgtm.

Issue #1593, opened 2016-08-21
Make sure \require{mhchem} doesn't override [Contrib]/mhchem


dpvc:
If the new `[Contrib]/mhchem/mhchem.js` is already loaded, then `\require{mhchem}` still loads the current `[MathJax]/extensions/TeX/mhchem.js` file, but it shouldn't, as the old version will replace the new one in this case.

The old file should check if `mhchem` is already loaded and not override it if it is.

dpvc:
The `issue1593` branch of my fork of MathJax includes a fix that prevents the old version from overriding the new one if it is already loaded.

dpvc:
==> In testsuite

LaTeXToMathML/mhchem/issue1593.html

dpvc:
==> Merged

PR #1596, opened 2016-08-21
Don't allow \require{mhchem} to override one already loaded.  #1593

dpvc:
Don't allow `\require{mhchem}` to override one already loaded from `[Contrib]`.

Resolves issue #1593.


zorkow:
lgtm.

Issue #1589, opened 2016-08-20
Bug: \ref and \eqref return the wrong URL if an HTML base element is specified in the document's head section


fizzbowen:
### Issue Summary
Specifying an HTML `base` element in a document's `head` section may cause the `formatURL` function in the `equationNumbers` object to return an incorrect URL to the equation.

### Steps to Reproduce

1. Create an HTML document which specifies the URL of a different document (e.g., the host site's home page) as the `href` attribute of the `base` element in the `head` section.
2. Invoke the CDN version of MathJax via `<script src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML" type="text/javascript"></script>` in the `head` section.
3. Use the default configuration, except enable auto-numbering by inserting `<script type="text/x-mathjax-config">MathJax.Hub.Config( { TeX: { equationNumbers: { autoNumber: "AMS"}}});</script>` into the `head` section.
4. Insert a numbered and labeled display equation into the document `body`, such as         `$$\begin{equation} \label{eq:general_linear_equation} {a_1}{x_1} + {a_2}{x_2} + {a_3}{x_3} + \ldots  + {a_n}{x_n} = b \end{equation}$$`.
5. Insert a reference to the equation elsewhere in the `body`, such as:
`A linear equation is "homogeneous" if and only if \(b = 0\) in Equation&nbsp;\( \eqref{eq:general_linear_equation}\).`
6. Upload the document to a Web server and open it in your favorite browser.
7. Mouse over the equation number (i.e., the "(1)") in the sentence where it is referenced to see the destination URL (or, if you are feeling brave, click the equation number to see where it takes you). Most likely, the destination URL will begin with the name of the document specified in the base element rather than the name of the document in which the equation appears.

This is a bug because...
...clicking on the referenced equation number should take you to that equation, not to an anchor in the base document.

Here is a fix which works for me (it restores the current document path to the URL)...
Locally changing the `formatURL` function in the `equationNumbers` object to read:
`formatURL: function (id) {return document.URL+'#'+escape(id)}`

However, I am not smart enough to know whether there would ever be a circumstance in which `document.URL` would also need to be escaped (as `id` currently is in the default version of `formatURL`), so this might be an additional consideration. And obviously making this the default would mess up references to an equation appearing in a different document.

### Technical details

* MathJax Version: master - latest commit:  Current CDN version as of Z 2016/08/20 19:00
[...]

PR #1602, opened 2016-08-24
Handle \ref properly when there is a <base> tag.  #1589

dpvc:
Handle references (`\ref` and `\eqref`) properly when there is a `<base>` tag.  Resolves issue #1589.


zorkow:
lgtm.

Issue #1584, opened 2016-08-16
MathJax.Hub.Queue fails when called from GWT function


kevinushey:
I'm running into trouble using MathJax within a GWT project. I have a GWT function using the native JSNI interface e.g.

```
    private static final native void mathjaxTypeset(String text, Element el) /*-{
      var MathJax = $wnd.MathJax;
      var f = function(x) { alert(x); };
      MathJax.Hub.Queue([f, 15]);
    }-*/;
```

Unfortunately, this fails for the reasons outlined by https://github.com/mathjax/MathJax/issues/508 -- we get a MathJax error of the form

```
<...> Can't make callback from given data
```

This ultimately stems from MathJax's use of `args instanceof Array` -- within the context of GWT, this check fails.

![screen shot 2016-08-16 at 3 46 01 pm](https://cloud.githubusercontent.com/assets/1976582/17718419/9ea732c8-63c8-11e6-8887-6ba110463385.png)

You'll notice that `instanceof Array` returns `false`, despite other checks reporting that this is indeed an array. Could this condition be relaxed to use a helper function e.g.

```
function isArray(object) {
    return Object.prototype.toString.call(object) === "[object Array]";
}
```

Or even `Array.isArray()` would suffice here.

Thanks!
Kevin

dpvc:
Thanks for your report, and proposed fix.  You are certainly correct that `instanceof` can be problematic, and certainly is in your case.  We ran into some similar issues in mathjax-node.

As you point out in your PR, we can't use `Array.isArray()` due to lack of support in some browsers, and MathJax has a policy against modifying global prototypes to add missing functionality like this.  Your suggestion of using `toString()` would work, but because some of these calls are in core pieces of MathJax (e.g., the Callback class, which is used very heavily), I would want to investigate the performance impact; I doubt it is significant, but that needs to be verified.

Also, your PR is a little more aggressive than necessary, as some of the instances you changes are not going to be in places where mixed origins for the arrays will occur, so `instanceof` is fine in those cases.  I would also want to make a utility function for this rather than reproduce the same code over and over.  So I'm going to close your PR for now, but appreciate the work you have done to try to overcome the issue and your willingness to share your solution.

@pkra, do you want to include something for this in v2.7?

kevinushey:
Thanks for your thoughtful response @dpvc -- I definitely agree a more nuanced version of my proposed PR would be more appropriate. Thanks for taking the time to respond + review!

dpvc:
[...]

PR #1603, opened 2016-08-24
Add MathJax.Object.isArray, and change most 'instanceof Array' to use it.  #1584

dpvc:
Add `MathJax.Object.isArray()` function, and change most `instanceof Array` to use it (unless the arrays are only going to generated by MathJax internals).  Also add `MathJax.Object.Array` for making arrays of the Array type used by MathJax.

Resolves issue #1584.

### DIFF
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -155,7 +155,7 @@ MathJax.cdnFileVersions = {};  // can be used to specify revisions for individua
     })
   });
   
-  BASE.Object.isArray = function (obj) {
+  BASE.Object.isArray = Array.isArray || function (obj) {
     return Object.prototype.toString.call(obj) === "[object Array]";
   };
   
--- a/unpacked/MathJax.js
+++ b/unpacked/MathJax.js
@@ -154,6 +154,12 @@ MathJax.cdnFileVersions = {};  // can be used to specify revisions for individua
 
     })
   });
+  
+  BASE.Object.isArray = function (obj) {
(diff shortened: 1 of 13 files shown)


unpacked/MathJax.js
```diff
@@ -154,6 +154,12 @@ MathJax.cdnFileVersions = {};  // can be used to specify revisions for individua
 
     })
   });
+  
+  BASE.Object.isArray = function (obj) {
+    return Object.prototype.toString.call(obj) === "[object Array]";
```
zorkow: I know it's a recommendation in the JS spec, but personally I always thought it is a poor way of checking for a type. But can't be helped.

zorkow:
lgtm.

unpacked/MathJax.js
dpvc: Agreed.  I preferred `instanceof Array`, but it does fail if the array comes from a different context (as was the case for the issue that spawned this, and is the case in mathjax-node, so this will reduce some problems with configuring MathJax in mathjax-node as well as resolve the original poster's problem.

dpvc: At least all the checking is done in one place now, so we can change it to something else in the future, if we want.  In fact, we could do

```
BASE.Object.isArray = (Array.isArray ? Array.isArray :
  function (obj) {...});
```

so that we only do the `toString()` test for older browsers.  Sorry I didn't think of that initially.

dpvc:
I updated to use `Array.isArray` when it is defined, and the Object `toString()` approach otherwise.

Issue #1605, opened 2016-08-25
update asciimath input for v2.7


pkra:
I think there were some relevant changes we should pull in.

dpvc:
The `issue1605` branch of my fork of MathJax includes the changes.  There seems to have been only [one commit](https://github.com/asciimath/asciimathml/commits/master/mathjax/jax.js) since 2.6.

pkra:
> There seems to have been only one commit since 2.6.

Yep, I noticed that as well.

@drlippman anything you want to fix before MathJax v2.7?

dpvc:
==> Merged.

drlippman:
There's been a few improvement requests I haven't gotten to yet. What's the timeline? None are a big deal, but might be nice to get them done if I can find the time.

pkra:
> What's the timeline?

Beta is starting pretty soon (this week or next) but I think we might be able to squeeze this in during the beta run.

drlippman:
OK, I'll see if there's anything I can get done in the next day or two.  Thanks.

drlippman:
I just made three to the asciimath repo with possible changes addressing lingering issues.  I'm not sure exactly what the process should be for vetting those changes.

The [bracket one](https://github.com/asciimath/asciimathml/pull/61) is really straightforward and can be merged without concern.  The other two have potential side effects, but I worry that no one else is going to look at them and give feedback.  It obviously would make things a lot easier if we had unit tests, but we don't :/

pkra:
Closing this as we finalizing the release.

PR #1606, opened 2016-08-26
Update AsciiMath to include changes since v2.6.  #1605

dpvc:
Update AsciiMath to include changes since v2.6.  Resolves issue #1605

### DIFF
--- a/unpacked/jax/input/AsciiMath/jax.js
+++ b/unpacked/jax/input/AsciiMath/jax.js
@@ -382,12 +382,13 @@ function createMmlNode(t,frag) {
 }
 
 function newcommand(oldstr,newstr) {
-  AMsymbols = AMsymbols.concat([{input:oldstr, tag:"mo", output:newstr, 
-                                 tex:null, ttype:DEFINITION}]);
-  // ####  Added from Version 2.0.1 #### //
-  AMsymbols.sort(compareNames);
-  for (i=0; i<AMsymbols.length; i++) AMnames[i] = AMsymbols[i].input;
-  // ####  End of Addition #### //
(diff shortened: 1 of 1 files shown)


pkra:
LGTM (in the sense that the changes are identical to the upstream commit).

unpacked/jax/input/AsciiMath/jax.js
```diff
@@ -666,14 +667,15 @@ function compareNames(s1,s2) {
...
 
 function initSymbols() {
-  var texsymbols = [], i;
-  for (i=0; i<AMsymbols.length; i++)
+  var i;
+  var symlen = AMsymbols.length;
+  for (i=0; i<symlen; i++) {
     if (AMsymbols[i].tex) {
-      texsymbols[texsymbols.length] = {input:AMsymbols[i].tex, 
+      AMsymbols.push({input:AMsymbols[i].tex,
```
zorkow: Bind AMsymbols[i] to an auxiliary variable.

pkra: @zorkow this code comes from upstream. Not sure how much we want to ensure code quality for upstream dependencies ;-)

zorkow: A PR comes in, I do a code review...
Anyway, if it's identical to upstream it's probably not worth it fixing here.

dpvc: Thanks for the code review.

I agree that it is not worth fixing here.  We have gone to great lengths *not* to modify the AsciiMathML.js code any more than necessary (so that keeping it up to date is easier).

Issue #1610, opened 2016-09-02
2.7 beta changes height for \| delimiter


cebola2:
2.7 beta noticeably shortens height for \\| delimiter without \left ot \right
2.6.1 seems to act more like what one would expect from TeX.

Screenshots are from https://cdi2tp.math.tecnico.ulisboa.pt/texto/14 See e.g. equation (1)

Version 2.6.1

![captura de ecra no 2016-09-02 10-17-54](https://cloud.githubusercontent.com/assets/1136663/18199159/0a2754c0-70f7-11e6-95ba-20d5ffdba349.png)

Version 2.7 beta

![captura de ecra no 2016-09-02 10-17-12](https://cloud.githubusercontent.com/assets/1136663/18199172/18bb08ce-70f7-11e6-804c-3ebf4b4f8481.png)

pkra:
Thanks for the report. I can confirm this. 

Only the TeX fonts seem affected.

dpvc:
The `issue1620` branch of my fork of MathJax includes a fix for this.

dpvc:
==> Merged

PR #1612, opened 2016-09-02
Use proper glyph for \| in TeX fonts (in HTML-CSS, CHTML, and SVG).  #1610

dpvc:
Remap U+2016 to U+2225 in normal and bold TeX fonts, now that `\|` produces U+2016 rather than U+2225, and U+2016 isn't in MAIN font.  Resolves issue #1610.


pkra:
All clear regarding CLA.

pkra:
Meh. wrong PR...

Issue #1670, opened 2016-12-02
Syntax error in cofiguration Javascript files, breaks IE 11


RichardSaffran:
Many or maybe all the Javascript files in the "config" directory of the downloaded MathJax installation contain the following syntax error: (line number varies, this is a searchable fragment)

     "border-spacing":0,}  // Invalid character: ,

That is not a problem with Mozilla Firefox but it breaks Javascript execution in Microsoft Internet Explorer 11.

dpvc:
The offending line is [line 85](https://github.com/mathjax/MathJax/blob/master/unpacked/jax/output/CommonHTML/jax.js#L84) of the CommonHTML renderer.  Note that this is not a problem for IE11 when used in IE8 emulation or above.  Note also that CommonHTML does not support IE7 or IE5 (i.e, Quirks) modes, so technically, it should not be a problem (though it should still be corrected).

If you are creating your own HTMLE pages, make sure that your file starts with `<!DOCTYPE html>` or has the proper `<meta>` tag in its head to force IE into a compatible emulation mode.  See [the documentation](http://docs.mathjax.org/en/latest/misc/faq.html?highlight=meta#what-should-ie-s-x-ua-compatible-meta-tag-be-set-to).  Use the Developer Tools (F12) to check that the document emulation is properly set.

dpvc:
The `issue1670` branch of my fork of MathJax includes the fix (but it doesn't include updated copies of the config file, since those aren't built until a release is being finalized).

RichardSaffran:
Wow, thanks for the quick response!

You wrote:

The offending line is line 85<https://github.com/mathjax/MathJax/blob/master/unpacked/jax/output/CommonHTML/jax.js#L84> of the CommonHTML renderer. Note that this is not a problem for IE11 when used in IE8 emulation or above. Note also that CommonHTML does not support IE7 or IE5 (i.e, Quirks) modes, so technically, it should not be a problem (though it should still be corrected).
I found that I have to add this to my files to get them to work with IE 11:

<meta http-equiv="X-UA-Compatible" content="IE=edge">

If I don’t do that, IE will either hang or display only the first letter of each equation (with CommonHTML), or the equations render in the wrong location (using HTML-CSS).

I haven’t tried an earlier version of IE.

--

dpvc:
> I found that I have to add this to my files

[...]

PR #1672, opened 2016-12-02
Remove unwanted trailing comma.  #1670

dpvc:
Remove unwanted trailing comma.  This causes IE < 8 to complain.  Even though CommonHTML doesn't support IE < 8, it is good not to leave such commas lying around.

 Resolves issue #1670.


RobLoach:
:+1:

zorkow:
lgtm.

PR #1740, opened 2017-04-24
Update for 2.7.1

dpvc:
This is the update to version 2.7.1.  It affects nearly all the files since they copyright and version numbers were updated.  It also includes the `latests.js` file, which allows loading the latest version from cdnjs, rawgit, or jsdelivr (whichever one it was loaded from).  It also includes the accessibility extensions in `extensions/a11y`, and the new mhchem extension in `extensions/TeX/mhchem3`, since these can be loaded automatically by the core MathJax.

I've smoke tested the release using rawgit.com, and all seems well.


pkra:
Could you update `package.json` to a matching version?

dpvc:
> Could you update `package.json` to a matching version?

Sorry, missed that.  Done.

pkra:
Using your branch via rawgit ([example](https://codepen.io/pkra/pen/qyzpf)), I run into 404s for many SRE json dependencies, in particular breaking the a11y extensions. 

I get fewer (but still some) 404s when loading the `/test` example from localhost.

pkra:
(PS also 404s for the audio files.)

pkra:
It seems `[MathJax`] does not get resolve and appears as a string in the paths, e.g.,

`GET http://localhost:8080/[MathJax]/extensions/a11y/mathmaps/functions/elementary.json 404 (Not Found)`

dpvc:
Yes, I was just looking into that.  Seems to be in the SRE code where it is using the path without resolving it.

I'll have to think about which is the best place to resolve that.

pkra:
Thanks, Davide!

dpvc:
OK, I think the place to fix it is [semantic-enrich line 102](https://github.com/mathjax/MathJax-a11y/blob/master/extensions/semantic-enrich.js#L102), which sets up the SRE path from the a11y path.  If it is changed to

```
  if (!PATH.SRE) PATH.SRE = MathJax.Ajax.fileURL(PATH.a11y);
```

then it will resolve any recursive paths before setting SRE, and so when SRE uses it to build its own paths, that will work out.  If someone sets SRE directly (the `if` check), then they would need to give an explicit path, or use `MathJax.Ajax.fileURL()` to resolve it when it is set, as is done here.

If that seems OK, I'll make an update to a11y.  If so, should that bump the version number of a11y?

pkra:
> If that seems OK, I'll make an update to a11y. 

:+1: 

> If so, should that bump the version number of a11y?

:+1. it's a bug fix, I'd say.

zorkow:
Sorry, whatever went wrong, is my fault. Peter asked me a couple of times
to look at the original PR and while I did have look, I never felt
comfortable to comment as I did not understand fully what was going on. (I
never commented either and I did misjudge the urgency...)

Let me test that first. I'll get back to you shortly.

Best,
Volker

On Mon, Apr 24, 2017 at 5:15 PM, Davide P. Cervone <notifications@github.com

zorkow:
OK, that works for me.
Volker

On Mon, Apr 24, 2017 at 5:15 PM, Davide P. Cervone <notifications@github.com

dpvc:
Thanks.  The issue is that in 2.7.1 the paths can no be nested, as in [a11y] points to [MathJax]/extensions/a11y, but SRE assumed (and rightly so given past experience) that the SRE path was absolute, it didn't take into account the possibility that there was another path being used recursively.

My solution was to make sure the SRE path is fully resolved when it is set.  That way, SRE can use it as it currently does.

I'll make the change in a11y, and then merge the PR immediately (since I'm taking this as the review).  Then I'll update the 2.7.1 branch of MathJax with the updated a11y files.

Davide

dpvc:
OK, I've updated 2.7.1-rc.1, and your example now works for me.  Thanks for catching that.  I had checked the collapsible extension, and thought that that used SRE, but apparently that wasn't enough.

pkra:
Awesome! 

All good from my end.

pkra:
Crap. I still get issue for the two audio files. @dpvc Can you take another look?

pkra:
Looks like https://github.com/mathjax/MathJax-a11y/blob/415ba3cc81f81f85b09e573d42c22be05a64e00f/extensions/explorer.js#L283

dpvc:
That looks right.  I'll make another update.

pkra:
Thanks, Davide!

dpvc:
OK, I've pushed another update.

pkra:
:+1: the smoke has cleared as far as I can tell. 

:shipit:

pkra:
Correction/clarification: "as far as I can tell" => "as far as I'm concerned"

dpvc:
I can do the merge, but will you do whatever is invalid in the release process?  I'm not sure I know what that is any longer.

Also, do you know what we need to do get this to cdnjs?  Does that happen automatically when there is an update to the MathJax master branch, or do we need to make a pull request?

pkra:
> I can do the merge, but will you do whatever is invalid in the release process?

(I'm assuming invalid was autocorrect for needed.)

Yes, I'll tag a release (with release notes) on GitHub; that's all we need now.

> Also, do you know what we need to do get this to cdnjs?

Happens automatically once there's a tag but I'll ping Peter to make sure.

PR #1750, opened 2017-05-19
Fix function name typo

ziktar:
laodDefaultMathJax => loadDefaultMathJax


dpvc:
Thanks for pointing out the error.  Our development process follows the "git-flow" model, where changes are branched from the `develop` branch, and pull requests are to the `develop` branch, so I'm going to close this PR since it doesn't follow that model.  I have made the corresponding change in a branch from `develop` as #1797, and cited you as the source of the fix.

PR #1756, opened 2017-05-30
Fix problem with namespaces on internal element.  Resolves #1755.

### LINKED ISSUE #1755: content-mathml.js and namespaces
Cf. https://github.com/mathjax/ctop/issues/34 

It currently breaks when content has namespace prefixes.

In particular, it cannot be used together with mml3.js

dpvc:
Remove namespaces of child elements as they are processed.  Fixes issue #1755.


zorkow:
lgtm.

PR #1779, opened 2017-06-29
Do not use “default” as attribute name, it is a reserved word in JS

mitya57:
In Debian we need to rebuild all minified JS files from their sources, and I use yui-compressor for that. It currently fails on this file which uses `foo.default` syntax.


pkra:
Thanks. This was already fixed upstream at https://github.com/mathjax/MathJax-a11y/pull/203 and will be in the next release.

mitya57:
Makes sense, thanks!

Issue #1694, opened 2017-01-25
newline in mtext produces "unknown character" warning [was: Newlines not properly handled inside `\text`, `\textrm` and similar macros]


jonasmalacofilho:
Depending inside where a newline appears, it may cause "unknown char" errors.

For instance, while

```tex
$$\textrm{foo 
bar}$$
```

[generates "SVG Jax - unknown char" messages](http://jsbin.com/gesodetuso/edit?html,console,output), a similar

```tex
$${foo 
bar)$$
```

works without warnings.

---

_MathJax: v2.7.0  
Client: Chrome 55/Node.JS v7.4.0_

pkra:
Thanks for catching that. It's only a warning; for the record, it happens on all outputs.

> Newlines not properly handled inside `\text`

There's merely a warning about the fonts not having a glyph for a character (and thus using a fallback from  a system font); the rendering is correct (and will be ok for most "unknown characters").

But this warning should be avoidable (i.e., strip the newline or handle it otherwise).

jonasmalacofilho:
Thanks! : )

Yes, the rendering was fine.  I only found out the warning when I enabled a more aggressive validation of all math in a project.

dpvc:
Until this gets fixed, here is a work-around for you.  You can add

```
<script type="text/x-mathjax-config">
  MathJax.Hub.Register.StartupHook("SVG Jax Ready",function () {
    MathJax.OutputJax.SVG.FONTDATA.REMAP[0x0A] = 0x20;
  });
  MathJax.Hub.Register.StartupHook("HTML-CSS Jax Ready",function () {
    MathJax.OutputJax["HTML-CSS"].FONTDATA.REMAP[0x0A] = 0x20;
  });
  MathJax.Hub.Register.StartupHook("CommonHTML Jax Ready",function () {
    MathJax.OutputJax.CommonHTML.FONTDATA.REMAP[0x0A] = 0x20;
  });
  MathJax.Hub.Register.MessageHook("HTML-CSS Jax - unknown char",function (msg) {
    console.log(msg);
  });
</script>
```

to your page just *before* the script that loads MathJax.js itself.  This will remap newlines to spaces in all three output jax, which should resolve the problem.

Hope that helps!

jonasmalacofilho:
That does help @dpvc, thanks!

I also came up with an example were the rendering was indeed affected:  `\textrm{x-\ny}` should render as `x- y` (with the space).

dpvc:
[...]

PR #1783, opened 2017-07-07
Remap newlines to spaces.  #1694.

dpvc:
Remap newlines to spaces, as is done in in TeX.  Resolves issue #1694.


unpacked/jax/output/SVG/fonts/TeX/fontdata.js
```diff
@@ -150,6 +150,7 @@
       RULECHAR: 0x2212,
       
       REMAP: {
+        0xA: 0x20,                      // carriage return
```
zorkow: Isn't CR 0xD? 0xA is LF (at least in ASCII).

dpvc: Oops.  I was originally doing both, but ended up only doing the line feed.  I got mixed up with the names when I removed the carriage returns.  I've fixed it.

Issue #1690, opened 2017-01-12
alignat: incorrect space before the equal sign [was: possible bug in alignat]


jason-s:
```
$$\begin{alignat}{2}
a &= \frac{1}{2}b+c+d &=& e        \\
f &= \frac{2}{3}g+h   &=& i+j+k
\end{alignat}$$ 
```

in https://cdn.mathjax.org/mathjax/latest/test/sample-dynamic-2.html renders like this:

![image](https://cloud.githubusercontent.com/assets/738893/21909738/6c001ce4-d8d6-11e6-8f01-97dfa4c178f5.png)

Note the small space before the `e` and `i`.

jason-s:
oh, could be user error on my part;

```
$$\begin{alignat}{2}
a &= \frac{1}{2}b+c+d &&= e        \\
f &= \frac{2}{3}g+h   &&= i+j+k
\end{alignat}$$ 
```

This looks better.

![image](https://cloud.githubusercontent.com/assets/738893/21909980/6e71d4f8-d8d7-11e6-95be-fa4451b5057b.png)

dpvc:
This is (almost) correct behavior for the `alignat` environment.  (In fact, there should be less space between the d and the equal sign.  This is actual LaTeX output:

<img width="411" alt="latex" src="https://cloud.githubusercontent.com/assets/432782/21966121/8c846e04-db3b-11e6-9334-87bfc258b17d.png">

The `alignat` environment is meant to produce columns of equations that are of the form

        a = b      c + c = d + d
    a + a = b + b      c = d

that is coded as

```
\begin{alignat}
        a &= b  &    c + c &= d + d \\
    a + a &= b + b  &    c &= d
\end{algnat}
```

Note that you do *not* put `&` after the equal signs.

If you are leaving out the left-hand side of an equality (like the `c + c` above), then you have to leave that column empty.  LaTeX makes special provisions for the equal signs to be properly spaced, but only does it every other column.  MathJax incorrectly modifies the spacing very *every* column, which is why there is space between the `d` and `=`.  Your extra `&` between `=` and `e` is why there *isn't* the correct space at that location.

dpvc:
Ive marked this "accepted" because of the incorrect space before the equal sign (which should only happen in even columns, not odd-numbered ones), not because of the missing space after the equal, which is correct behavior.

dpvc:
The `issue1690` branch of my fork of MathJax includes a fix for this.

dpvc:
[...]

PR #1784, opened 2017-07-07
Fix spacing aligned/alignedat.  #1690.

dpvc:
The aligned/alignedat environment has to adjust spacing around equal signs and other relations (TeX spacing requires an operand on both sides to get the right spacing, so we add a blank one when needed).  This should only occur every other column, however.  This corrects the problem the MathJax was adding the empty elements on *every* column but the first one.

Resolves issue #1690.

### DIFF
--- a/unpacked/extensions/TeX/AMSmath.js
+++ b/unpacked/extensions/TeX/AMSmath.js
@@ -549,7 +549,7 @@ MathJax.Hub.Register.StartupHook("TeX Jax Ready",function () {
       stack.global.tagged = !numbered && !stack.global.forcetag; // prevent automatic tagging in starred environments
     },
     EndEntry: function () {
-      if (this.row.length) {this.fixInitialMO(this.data)}
+      if (this.row.length % 2 === 1) {this.fixInitialMO(this.data)}
       this.row.push(MML.mtd.apply(MML,this.data));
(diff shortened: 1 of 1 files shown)


unpacked/extensions/TeX/AMSmath.js
```diff
@@ -549,7 +549,7 @@ MathJax.Hub.Register.StartupHook("TeX Jax Ready",function () {
       stack.global.tagged = !numbered && !stack.global.forcetag; // prevent automatic tagging in starred environments
     },
     EndEntry: function () {
-      if (this.row.length) {this.fixInitialMO(this.data)}
+      if (this.row.length % 2 === 1) {this.fixInitialMO(this.data)}
```
zorkow: Maybe I misunderstand the function: But this looks like it changes spacing only if the row is of an odd length.
Should it not change spacing on every odd column, regardless of the length of the row?

dpvc: The row is being built at this point (we are adding the next cell to it), so the row length is the current cell count.  It is not the complete row, but the partial row at this point.

zorkow: Thanks. Makes sense.

Issue #1735, opened 2017-04-14
[CommonHTML] Poor over script position above italic letters


samdooley:
### Issue Summary

Over script characters that are positioned above italic characters appear too far to the right, as if the italic correction is not handled properly.

### Steps to Reproduce

1.  Load the following HTML example page.

```
<script type="text/javascript"
   src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=MML_CHTML">
</script>

<math xmlns="http://www.w3.org/1998/Math/MathML" mathsize="36pt">
  <mover>
    <mrow>
      <mi>X</mi>
      <mo>&#x2063;</mo>
      <mi>J</mi>
    </mrow>
    <mo>&#x23dc;</mo>
  </mover>
  <mo>=</mo>
  <mover>
    <mrow>
      <mtext mathvariant="italic">X</mtext>
      <mo>&#x2063;</mo>
      <mtext mathvariant="italic">J</mtext>
    </mrow>
    <mo>&#x23dc;</mo>
  </mover>
</math>
```

2.  Observe the horizontal position of the over script character in relation to the italic base letters.

This is a bug because the extent of the over script character should match the extent of the base letters, after accounting for the italic slant of the characters.  The over script extends too far to the right, especially for the case on the left that uses mi elements.

### Technical details

* MathJax Version: 2.7.0 release
* Client OS: Windows
* Browser: Chrome

![mover-italic](https://cloud.githubusercontent.com/assets/4752336/25056581/13ec0560-2127-11e7-941c-c5efa136fee1.png)

pkra:
Thanks for the report. I can reproduce this with the CommonHTML output on Chrome and Firefox.

dpvc:
I believe the problem is [this line](https://github.com/mathjax/MathJax/blob/master/unpacked/jax/output/CommonHTML/jax.js#L1418), which bubbles the accent skew up from lower nodes to higher nodes.  The skew value is only supposed to be used if base is a single character, but this propagates the skew value for all cases.  I think

```
            if (cbox.skew && this.data.length === 1) bbox.skew = cbox.skew;
```

would do the trick.

samdooley:
Thanks for the patch.  I was able to test it with a local 2.7.0 and it is better, but I believe the line above needs the same treatment.  So I think:

            if (cbox.ic && this.data.length === 1) {bbox.ic = cbox.ic} else {delete bbox.ic}
[...]

PR #1787, opened 2017-07-07
Improve position of stretchy accents.  #1735

dpvc:
Only bubble up italic correction or skew for nodes with only one child.

Resolves issue #1735

### DIFF
--- a/unpacked/jax/output/CommonHTML/jax.js
+++ b/unpacked/jax/output/CommonHTML/jax.js
@@ -1415,8 +1415,13 @@
           if (!options.noBBox) {
             var bbox = this.CHTML, cbox = child.CHTML;
             bbox.append(cbox);
-            if (cbox.ic) {bbox.ic = cbox.ic} else {delete bbox.ic}
-            if (cbox.skew) bbox.skew = cbox.skew;
+            if (this.data.length === 1) {
+              if (cbox.ic) bbox.ic = cbox.ic;
+              if (cbox.skew) bbox.skew = cbox.skew;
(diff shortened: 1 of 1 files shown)


zorkow:
I've been playing around with the example from the issue and observe some very bizarre behaviour:
* in HTML-CSS no accent is displayed
* in CommonHTML the enriched math becomes huge
* in SVG it looks good
Maybe we can discuss this tomorrow.

dpvc:
Both issues seem to be related to the `mathsize="36pt"` specification, and have nothing really to do with the issue being addressed here.  I think they should be added as a new issue instead of holding up this one.  (Note that this issue is about horizontal spacing within CHTML only.)

dpvc:
The HTML-CSS issue is due to the way HTML-CSS positions the accents, which fails for large `mathsize` settings.  I've opened #1798 for this.

zorkow:
Sounds good.

zorkow:
CommonHTML problem could be an maction issue. It's independent of accents. I'll file an issue.

dpvc:
I'm working on it already.  It is more than just mactions, it is the collapsible extension itself, somehow.  Just starting to look.

zorkow:
I've filed it in #1800. Feel free to close or move to a11y.

dpvc:
Thanks.  I'm working on it and will figure out if it is a CHTML or an a11y-based problem (or both).  It does seem to only affect CHTML, but that might be just a difference in how CHTML handles `mathsize`.

PR #1791, opened 2017-07-07
Fix CD problems cause by getNode().  #1691

### LINKED ISSUE #1691: [CommonHTML] AMScd error
Hello,

In a previous version of MathJax (which I do not remember), the following AMScd diagram was rendered correctly: 
```
\begin{CD}
A @>a>b> B \\
@VlVrV @AlArA \\
C @<a<b< D
\end{CD}
```
but it is not the case in the current version.
Could you please check?
Thank you in advance.

dpvc:
Some situations caused CD to produce Math Processing Errors (see #1691 for reduced test case).  This was caused by the `getNode()` method, which did not descend into all the children.  This PR fixes the problem by using a breadth-first search in `getNode()`, but stopping at nodes with `id`s, as these are other rendered MathML nodes (we only want to find nodes that are "direct" children of the current element, not similarly named ones inside nested elements).

Resolves #1691.

### DIFF
--- a/unpacked/jax/output/CommonHTML/jax.js
+++ b/unpacked/jax/output/CommonHTML/jax.js
@@ -376,12 +376,16 @@
     //
     getNode: function (node,type) {
       var name = RegExp("\\b"+type+"\\b");
+      var nodes = [];
       while (node) {
         for (var i = 0, m = node.childNodes.length; i < m; i++) {
           var child = node.childNodes[i];
-          if (name.test(child.className)) return child;
+          if (child) {
+            if (name.test(child.className)) return child;
+            if ((child.id || "") === "") nodes.push(child);
+          }
         }
-        node = (node.firstChild && (node.firstChild.id||"") === "" ? node.firstChild : null);
+        node = nodes.shift();
       }
       return null;
     },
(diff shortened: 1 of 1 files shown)


unpacked/jax/output/CommonHTML/jax.js
```diff
@@ -376,12 +376,16 @@
...
     getNode: function (node,type) {
       var name = RegExp("\\b"+type+"\\b");
+      var nodes = [];
       while (node) {
         for (var i = 0, m = node.childNodes.length; i < m; i++) {
           var child = node.childNodes[i];
-          if (name.test(child.className)) return child;
+          if (child) {
+            if (name.test(child.className)) return child;
+            if ((child.id || "") === "") nodes.push(child);
```
zorkow: Why is this not just !child.id ? !"" is true. Or is that some old IE legacy I am not aware of?

dpvc: I was trying to avoid the case where `id` is not defined, but it turns out it always is (as far as I can tell).  We should be able to just do `child.is === ""` in that case.  I'll change it.

dpvc: Done.

Issue #1686, opened 2017-01-09
[HTML-CSS, SVG] \underline misplaced in scriptstyle


dpvc:
In HTML-CSS and SVG output, the `\underline` macro does not place the underline properly.  For example:

    \scriptscriptstyle \underline{k}

shows no underline in HTML-CSS and misplaced underline in SVG output.  CommonHTML shows the underline, but it is rather low.

This is originally from [math.SE](http://meta.math.stackexchange.com/questions/25614/underline-does-not-display-in-fraction-mathjax-bug).

dpvc:
The `issue1686` branch of my fork of MathJax includes a fix for this.

dpvc:
==> Merged

dpvc:
==> In testsuite

MathMLToDisplay/Presentation/ScriptsAndLimits/mover/issue1686.html

PR #1792, opened 2017-07-07
Fix placement of underlines in script styles.  #1686

dpvc:
Since 8f628e0590f87f073185c1e7938964dbb1c41002 changed the height and depth of the minus sign to match that of the plus sign (to match TeX's requirements for its usage), the use of U+2212 as a stretchy horizontal character is now causing alignment issues in some cases.  This PR switches U+2013 (en-dash) for U+2212 in stretchy characters to avoid that problem in the MathJax TeX fonts.

Resolves issue #1686

### DIFF
--- a/unpacked/jax/output/CommonHTML/fonts/TeX/fontdata.js
+++ b/unpacked/jax/output/CommonHTML/fonts/TeX/fontdata.js
@@ -389,6 +389,10 @@
       {
         dir: H, HW: [[.333+.25,MAIN],[.555+.25,SIZE1],[1+.33,SIZE2],[1.443+.33,SIZE3],[1.887,SIZE4]]
       },
+      0x2013: // en-dash
+      {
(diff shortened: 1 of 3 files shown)


unpacked/jax/output/CommonHTML/fonts/TeX/fontdata.js
```diff
@@ -167,8 +167,6 @@
       {name: "greek", low: 0x03B1, high: 0x03F6, offset: "G"}
     ],
       
-    RULECHAR: 0x2212,
```
zorkow: Is RULECHAR not being used anymore? Or should that be 0x2013?

dpvc: Correct, it is not used, and never was in CHTML.  It was copied from the HTML-CSS data when I was first writing CHTML, but was never used, and I didn't catch it until now.  The second commit above indicated that it wasn't used in the commit message, but I should have pointed that out in the PR message.

Issue #1482, opened 2016-05-19
Handle \text in \cases better


dpvc:
As reported in this [meta.math.SE post](http://meta.math.stackexchange.com/questions/23241/tex-equations-not-rendering-the-not-equal-symbol-correctly), the backwards-compatibility check used to detect when `\text{}` is already in the second column is not robust enough, and something like

```
$$d_{1}\left(x,y\right)=\cases{
  0&\text{if }x=y\cr
  1&\text{if }x\neq y}$$
```

causes the check to put the second column in math-mode when it shouldn't.

dpvc:
The `issue1482` branch of my fork of MathJax includes a fix for this.

dpvc:
==> In testsuite

LaTeXToMathML/issue1482.html

dpvc:
==> Merged

PR #1793, opened 2017-07-07
Better test for \text{} in \cases{}.  #1482

dpvc:
Better test for `\text{}` in `\cases{}` macro.  This was for backward compatibility, but it wasn't working quite right.  This should handle the situation properly by tracking the closing brace of the `\text{}` macro, and checking that it is at the end of the string.

Resolves issue #1482.


dpvc:
Do you mean in the code, or here?

zorkow:
In the code. There's so many cases.

dpvc:
I've added the comments.

Issue #1709, opened 2017-03-01
[CommonHTML] Safari extraneous space in mover 


pkra:
[Reported via the user group](https://groups.google.com/forum/#!topic/mathjax-users/J9-jy6_tVmM).

Using the CommonHTML output, Safari 9 and 10 add extra space to the right of an `mover`; HTML-CSS and SVG seem ok. This might be a browser bug.

Example.
```
<math xmlns="http://www.w3.org/1998/Math/MathML">
  <mrow class="MJX-TeXAtom-ORD">
    <mover>
      <mi>u</mi>
      <mo stretchy="false">&#x2192;<!-- → --></mo>
    </mover>
  </mrow>
  <mo>+</mo>
  <mrow class="MJX-TeXAtom-ORD">
    <mover>
      <mi>v</mi>
      <mo stretchy="false">&#x2192;<!-- → --></mo>
    </mover>
  </mrow>
</math>
```

dpvc:
The vector arrow is one of the combining characters, so it is zero-width.  The combining characters have been a perennial headache, as the browsers handle them differently, and there is often problems when they are used without a preceding character (as is the case for MathJax, which positions them by hand).  It looks like Safari has made a change that affects this (they are incorrectly handing the zero width, and it is getting a non-zero width).  It looks like explicitly setting `width:0` on the parent element will do the trick.

pkra:
>  It looks like Safari has made a change that affects this (they are incorrectly handing the zero width, and it is getting a non-zero width)

I'll try to make file a bug upstream.

dpvc:
Well, I suspect that Safari's behavior is by design, and will not be considered a bug by Apple.  The combining characters like the vector accent (U+20D7) are implemented by making the character be zero-width and overlapping on the left, so that when it follows a character, it extends over that character.  Most browsers also may add a vertical shift to accommodate tall letters (the accent is usually positions in the font to cover a lower-case letter like "x").  Some browsers will also try to center the accent over the preceding character (e.g., Chrome).

Since these characters are intended to combine with the previous character, the expected behavior for when there *is* no preceding character is not clear.Some browsers (e.g., IE) would not show the character at all, for example, but until Safari 10, most other browsers simply showed the character to the left of the starting position (as though there actually *were* something to overlap.  As of Safari 10, however, when there is no preceding character, the combining character is instead treated like a *non*-combining character, and so it extends to the right of the starting point and takes up non-zero space.

[...]

PR #1796, opened 2017-07-07
Use non-combining vector arrow for \vec.  #1709

dpvc:
This PR works around the changes in how combining-characters are handled in WebKit >= 603 (they are treated as non-combining when there is no character directly preceding them).  We include a new single-character font with a non-combining version of this glyph, so all browsers can treat this character the same, and remap U+20D7 to that character when used as an accent.  (We have to go through a temporary fake character since `REMAPACCENT` doesn't allow specifying the variant.)

Resolves issue #1709.


dpvc:
The alternative would be to extend the `remap()` method used by the accent code to allow the data to specify a variant as well as a remapped character, but that introduces complexity that I'd rather avoid.  There are several other hacks in the font data.  I'm hoping that that will all be straightened out in the new v3 font handling mechanisms.

dpvc:
Did you want me to do it a different way?

zorkow:
No, don't do it differently. 
I was just referring to you comment that in the code that it is a hack. And I had to agree.
But please merge.

Issue #1806, opened 2017-07-12
Using 'const' keyword with certain global variables causes mathjax to throw exceptions on render


applmak:
Here's a simple repro case:

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_CHTML"></script>
<script>
const scale = 1.0;
</script>
\(x = 1\)
```

I would expect this to render `x = 1`, but instead, I see `[Math Processing Error]`.

dpvc:
Thanks!   Looks like a non-localized variable used in a function.  Argh!  I thought we had tested for that, but apparently not.  I'll fix the reference, and look into why this was missed...

dpvc:
The `issue1806` branch includes the fix for this.  I found one other variable that was also global that should not have been.  The compressor usually reports these, but for these cases it doesn't.  Not sure why.

dpvc:
==> In testsuite

UI/issue1806.html

dpvc:
==> Merged

PR #1809, opened 2017-07-14
Localize some variables that where accidentally global.  #1806

dpvc:
Localize some variables that where accidentally global in CommonHTML.

 Resolves issue #1806.

### DIFF
--- a/unpacked/jax/output/CommonHTML/autoload/mtable.js
+++ b/unpacked/jax/output/CommonHTML/autoload/mtable.js
@@ -220,7 +220,7 @@ MathJax.Hub.Register.StartupHook("CommonHTML Jax Ready",function () {
         for (var j = 0, M = row.length; j < M; j++) {
           var s = (rdata.type === "mtr" ? 0 : LABEL);
           var mtd = rdata.data[j-s] || {CHTML: CHTML.BBOX.zero()};
-          cell = row[j].style; cbox = mtd.CHTML;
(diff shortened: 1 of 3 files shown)


TomasRiker:
Unbelievable coincidence ... I wanted to see whether the `scale` variable did indeed leak into the global object and I found a bug in my own website that also leaked a variable of the same name.
An idea for automated testing of such things: Make a snapshot of all global variables, then load MathJax, display a bunch of different math content, then make a snapshot again and check that there are no new unexpected global variables.

dpvc:
Yes, that is how I found these two (after one was reported in #1806).  The compressor we use reports these variables in some cases (and I thought all cases), but apparently not.  So something like what you suggest would help.  But MathJax consists of 10's of thousands of line of code, and unless you run all of it, you aren't guaranteed to find all global variables.  But it would at least be a step in the right direction.

TomasRiker:
Assuming that the regular tests cover most of MathJax, would it be an option to do the check before/after running all tests in a row, or before/after each individual test?

dpvc:
That is a nice idea, but many of the tests create globals as part of the test, so they would cause problems for this approach.  It would require modifying many of the test files to make that work.

The test framework for version 3 will be entirely rewritten, so we don't expect to put a lot of effort into the current test suite at this point.  Also, version 3 is using Typescript, which will catch these sorts of errors in the development phase.

PR #1723, opened 2017-03-21
Add all macros currently supported by MediaWiki

physikerwelt:
Add all macros definitions that are currently available on
Wikipedia and related projects. Those additional
symbols were introduced in

https://github.com/wikimedia/MathJax/commit/ff2accb8a2dfa5ad03fbd01db52dd1e12c39991e

cf. https://phabricator.wikimedia.org/T105310

PR #1810, opened 2017-07-14
Add more unicode-based intergrals.  See PR #1723.

dpvc:
This defines the additional unicode-based integrals requested in PR #1723, but does so in a way that will scale for displaystyle.  They will act like the usual integrals except that the position of the limits of integration will not include the italic correction, so they won't be staggered as they would be with `\int`.  That would also be true of the versions in the original PR, unless the STIX fonts were used (since these characters are available in STIX, and so the font metrics are available there).  But it appears that the MediaWiki SVG files use the MathJax fonts, not the STIX ones, so the versions here make more sense.  If they used STIX, then the originals would be better, as they would properly scale and use the correct limit positions automatically.


pkra:
cc @physikerwelt

Issue #1813, opened 2017-07-27
[2.7.2-beta.0] There is still a timing issue with accessibility-menu in combined config files


dpvc:
If the explorer is active when the page is loaded (due to the menu item being checked from an earlier season on that site), then the speech generation may not be triggered.  Selecting an expression and pressing shift-space brings up the menu rather than starts the explorer.

This only happens with the combined configuration files in the packed directory (not the unpacked one).  The issue is due to the fact that in the packed combined configuration file, the extension is not in the extensions array, and so the loading of extensions doesn't wait until the accessibility-menu extension is fully loaded.  So the math is typeset before the accessibility-menu loads the explorer extension, and the speech key bindings are not added to the math since they are already typeset.

The solution is to set the extensions array to include the accessibility-menu anyway, even though it is loaded by the combined configuration file, as that will make sure that it blocks the extension code from completing until any needed accessibility extensions are loaded.

dpvc:
Resolved as part of the recombining for beta.1

PR #1814, opened 2017-07-29
Fix problem with timing of accessibility-menu.  #1813

dpvc:
Add accessibility-menu.js to extensions array to get timing right.  Resolves issue #1813.


dpvc:
This is obsolete since the changes are included in the recombining for beta.1

PR #1878, opened 2017-12-02
fix `HUB.Startup.signal.Post()` queued invocations: API expects STRING param, received ARRAY instead

GerHobbelt:
fix: `HUB.Startup.signal.Post()` API expects a single message STRING, so we must concatenate parts into a single string before queueing a call to this API. Fixes *crash* in `HUB.Startup.signal.Post()` itself where code expects a string type call parameter, rather than an array type parameter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mathjax",
-  "version": "2.7.2-rc",
+  "version": "2.7.2",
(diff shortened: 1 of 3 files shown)


pkra:
Please follow https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md#key-branches--tags and only make PRs to the develop branch.

pkra:
Closing this to avoid accidental merges.

PR #1879, opened 2017-12-05
Typo: fix 'initial' typo

luzpaz:
Fixes what seems like a functional typo. Please review. Found via downstream scipy project.


dpvc:
Thanks for catching the typo.  Most of the files you have changed are generated automatically from the one in the unpacked directory, so that is the only file that needs changed.  Because of that, I'm going to close this pull request (but that doesn't mean we don't value the contribution).

I will make a new PR for the one file that needs to be edited.

PR #1880, opened 2017-12-05
Misc. trivial typos

luzpaz:
Most are source comments. Few are user-facing. Found using  `codespell -i 3 -w  --skip="./localization,./unpacked/localization" -I ../mathjax-whitelist.txt` 
(typos originally identified downstream in scipy)


dpvc:
Thanks for catching all of these.  It is embarrassing to see them all.

The font files are generated programmatically (and the program had the typo), so I have fixed it there, in case they are generated again in the future.  As with your other pull request, only the unpacked files are necessary, but because of the large number of font files adjusted, I'll merge this one.

PR #1889, opened 2017-12-26
fix `HUB.Startup.signal.Post()` queued invocations: API expects STRING param, received ARRAY instead

GerHobbelt:
This is a redo of #1878 -- trouble there was I _had_ prepped the patch against `develop` branch but had picked the `master` branch in the GitHub 'Open New PullReq' UI. 😞 

---

fix: `HUB.Startup.signal.Post()` API expects a single message STRING, so we must concatenate parts into a single string before queueing a call to this API. Fixes crash in `HUB.Startup.signal.Post()` itself where code expects a string type call parameter, rather than an array type parameter.


dpvc:
Thanks for catching that.

PR #1877, opened 2017-12-02
fix: typos in comments and a message string (English)

GerHobbelt:
See commit/subject line. Minimal edit.

PR #1890, opened 2017-12-26
fix: typos in comments and a message string (English)

GerHobbelt:
redo of #1877, which went wrong due to same mistake as mentioned in #1889.


dpvc:
Thanks for catching those.

PR #1893, opened 2017-12-28
Do not use let as variable name

mitya57:
It produces an error with yui-compressor 2.4.8 on Debian.


dpvc:
Thanks for the contribution.

PR #1924, opened 2018-02-07
Adds new version of the a11y extension.

zorkow:
* SRE 2.2.1
* Streamlined Accessibility menu
* Adds mathmaps for Spanish locale


dpvc:
The unpacked copies still need to be moved to the `unpacked/extensions/a11y` directory, but I'll take care of that.

PR #2029, opened 2018-07-11
Add extension for LaTeX package braket

wgertler:
"braket" is a light LaTeX package that provides macros for typesetting the notational conventions of quantum mechanics


pkra:
Have you checked https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md ? You're not following the folder structure of MathJax and you're working on the wrong branch.

Overall, this looks like you wre trying to make a third party extension, cf. https://docs.mathjax.org/en/latest/options/ThirdParty.html#thirdparty. In that case, it's usually best to start out in a separate repository and publish to npm for distribution (also via CDNs like jsdelivr).

zorkow:
I note that the braket and set macros differ from the LaTeX package, in that they both take two arguments, while the original LaTeX macros only take one argument and use | or || as separators.
Is that intended? I would advise changing the arity of macro in MathJax.

Also in the Braket macro the `\middle` in front of the `\mid` seems to be missing.

wgertler:
@zorkow you're right, thanks for the assist. I'm going to make the changes you suggested- they're more correctly in-line with the actual braket TeX package, where as what I've got here is really just personal preference. Doesn't make sense to call it support if the macros aren't the same here as in TeX.

@pkra Thanks for the feedback, I'll be more careful and try to make the right changes, and publish to npm. 

Cheers!

zorkow:
I've put together a quick implementation of the braket package (with the expected behaviour) in version 3.
Have a look [here](https://github.com/mathjax/mathjax-v3/tree/braket/mathjax3-ts/input/tex/braket).
It's still in its own branch, but should make it into the next beta release.

JeppeKlitgaard:
I think this PR can be closed now, as a `braket` implementation now exists in MathJax v3+ using the `braket` extension?

Issue #2035, opened 2018-07-25
[STIX-web] bold-italic glyphs no longer found (in 2.7.5)


pkra:
E.g., `<math><mi mathvariant="bold-italic">x</mi></math>` fails to locate a glyph in the STIX-web fonts (in both SVG and HTML-CSS).

Possibly a regression after #1923 (since 2.7.4 uses STIXWEBMAINBI).

dpvc:
Argh!  I used the wrong offset for the bold-italic letters.  You can fix this using

```
MathJax.Hub.Register.StartupHook("HTML-CSS Jax Ready", function () {
  var HTMLCSS = MathJax.OutputJax["HTML-CSS"];
  HTMLCSS.Augment({FONTDATA: {VARIANT: {'bold-italic': {offsetA: 0x1D468}}}});
});
MathJax.Hub.Register.StartupHook("SVG Jax Ready", function () {
  var SVG = MathJax.OutputJax.SVG;
  SVG.Augment({FONTDATA: {VARIANT: {'bold-italic': {offsetA: 0x1D468}}}});
});
```

which will set up the correct values.  You should be able to put these into a local extension that you load into mathjax-node via the `--extensions` option until the data is fixed officially.

pkra:
Thanks, Davide!

pkra:
@dpvc I recently ran into the same problem with bold italic greek characters reported as unknown characters. I believe STIX 1 has those as well.  If you could share a similar patch, that'd be great.

dpvc:
@pkra, I'm not able to reproduce that.  With

```
#! /usr/bin/env node

var mjAPI = require("./lib/main.js");

mjAPI.config({MathJax: {SVG: {font: "STIX-Web"}}});

mjAPI.typeset({
  math: '<math><mtext mathvariant="bold-italic">&#x393;&#x3B3;</mtext></math>',
  format: "MathML",
  svg: true,
  ex: 6, width: 100,
}, function (data) {
  if (!data.errors) {
    console.log(data.svg);
  }
});
```

I get the output of

```
<svg xmlns:xlink="http://www.w3.org/1999/xlink" width="2.461ex" height="2.509ex" style="vertical-align: -0.671ex; margin-left: -0.029ex; margin-right: -0.053ex;" viewBox="-12.5 -791.3 1059.5 1080.4" role="img" focusable="false" xmlns="http://www.w3.org/2000/svg" aria-labelledby="MathJax-SVG-1-Title">
[...]

PR #2038, opened 2018-07-27
Correct offset for bold-italic in STIX font data.  #2035.

dpvc:
Correct offset for bold-italic in STIX font data. 

Resolves issue #2035.


zorkow:
Sorry, wrong repo...

PR #2042, opened 2018-08-05
Support TeX-in-<math>-tags, a la Wikipedia's math markdown.

### LINKED ISSUE #2043: Support TeX-in-<math>-tags, a la Wikipedia's math markdown
I don't trust myself never to write `$$` or `\[` by accident (in a non-math context) on my blog, so I'd like to see MathJax support Wikipedia-style markdown for math:

    Inline as <math>1+2=3</math> or out-of-line as
    <math mode=display>
        1+2 = 3
    </math>

The docs for `tex2jax` indicate that `tex2jax` cannot handle delimiters with angle brackets in them, and I've verified that that's the case. However, `mml2jax` can handle `<math>` tags just fine, and I have managed to make a patch that implements the above-requested behavior.

Quuxplusone:
This minor change in `mml2jax` allows you to write inline math markup just like you would on Wikipedia or Stack Overflow: `<math>1+2=3</math>`. We accomplish this by mixing the "mml2jax" preprocessor (which knows how to find math tags) with the "tex" middle-end (which knows how to render TeX input).

The one complication is that the TeX middle-end expects to see the text *without* the surrounding `<math> ... </math>`, whereas the MathML middle-end expects to see it *with* the surrounding `<math> ... </math>`. So if we're outputting to the TeX middle-end, we need to take `innerHTML` instead of `outerHTML`.

To test the functionality, copy `mml2jax.js` from `unpacked/extensions/` into `extensions/`, and then open `index.html` in a Web browser. You should see two identical display-mode equations: one rendered by `<math mode=display> ... </math>` and one rendered by `\[ ... \]`.

I plan to use this extension on my blog, which uses Jekyll, and where I don't trust myself to avoid writing `\(` or `\[` or `$$` by accident. Whereas I definitely do trust myself to never write `<math>` by accident!

Fixes #2043.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -16,8 +16,15 @@
 //
   MathJax.HTML.Cookie.Set("menu",{});
   MathJax.Hub.Config({
-    extensions: ["tex2jax.js"],
+    extensions: ["mml2jax.js", "tex2jax.js"],
     jax: ["input/TeX","output/HTML-CSS"],
+    mml2jax: {
+      outputTex: true
+    },
+    tex2jax: {
+      inlineMath: [['\\(', '\\)']],
+      displayMath: [['$$', '$$'], ['\\[', '\\]']],
+    },
     "HTML-CSS": {
       availableFonts:[],
       styles: {".MathJax_Preview": {visibility: "hidden"}}
@@ -98,6 +105,12 @@ TeX code instead, MathJax is not working for you.
 <!------------------------------------------------------------------------>
 <hr>
 
+
+<p>
+<math mode="display">
+\frac{-b\pm\sqrt{b^2-4ac}}{2a}
+</math>
+</p>
 
 <p>
 \[
--- a/unpacked/extensions/mml2jax.js
+++ b/unpacked/extensions/mml2jax.js
@@ -29,14 +29,15 @@
 MathJax.Extension.mml2jax = {
   version: "2.7.5",
   config: {
-    preview: "mathml"       // Use the <math> element as the
+    preview: "mathml",      // Use the <math> element as the
                             //   preview.  Set to "none" for no preview,
                             //   set to "alttext" to use the alttext attribute
                             //   of the <math> element, set to "altimg" to use
                             //   an image described by the altimg* attributes
                             //   or set to an array specifying an HTML snippet
                             //   to use a fixed preview for all math
-
+    outputTex: false        // Produce "math/tex" script tags instead of
+                            //   "math/mml" script tags.
   },
   MMLnamespace: "http://www.w3.org/1998/Math/MathML",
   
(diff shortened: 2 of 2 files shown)


dpvc:
First, thank you for your interest in MathJax, and your willingness to contribute to the project.  We appreciate that very much.  There are several reasons, however, why we are going to reject this pull request. 

The first is that the `mml2jax` preprocessor is intended to find MathML in the page while `tex2jax` is for finding TeX in the page (as you already know).  Adding your changes conflates these two separate functions in a way that muddies the purposes of the various components.

Second, in HTML5, the `<math>` tag has a well-defined meaning:  it introduces MathML (not TeX).  Since the page will be parsed by the browsers ahead of MathJax running, the browsers will try to interpret your TeX as MathML (and that should produce errors, since it is not valid MathML).  Wikipedia's markdown is not HTML, but instead is processed on their servers to generate the final HTML.  They can use `<math>` in whatever way they wish, but the final HTML page they generate does not include `<math>` tags (at least not ones that contain TeX code).  Personally, I think it is a bad choice for them to have used `<math>` as their delimiters, but it is their choice how to structure their derivative of Markdown.  But in an HTML page, you don't have that luxury, as `<math>` already means MathML there.

Third, the `mml2jax` preprocessor looks for a variety of forms for the `<math>` tags, including tags with namespaces, like `<m:math>`, and so on.  Your code would process those as well.

Fourth, since you are using `innerHTML` to obtain the TeX code, if the TeX included HTML special characters like `<`, `>`, or `&`, these will be translated into entities, and you would get `&lt;`, `&gt;`, and `&amp;` in your TeX input instead.  These will cause errors in processing the TeX, and won't produce the desired expressions.  (You probably want to use `textContent` rather than `innerHTML`.)

Fifth, because HTML tags are allowed inside `<math>` tags, you would potentially be allowing math that contains HTML tags to be processed (though they would come through as literal `<` and `>` around the tag names).  MathJax does not allow HTML tags (other than `<br>`) inside the math it processes, so this is inconsistent with `tex2jax`, and it makes not real sense to treat the tags as literal strings.  Even using `textContent` would basically ignore the presence of tags, which is counter-intuitive.

---

If you want to use `<math>` rather than `\(` or other textual delimiters, then what you probably should do is make a new preprocessor for that, rather than corrupting the purpose of an existing one that does something else.  You can start with `mml2jax` as a basis, but remove the MathML handling, and replace it with your own TeX handling.  Of course, you would rename it to something else as well.

You can publish your new preprocessor as a third-party extension to MathJax, like the current `mhchem` and arabic extensions (which are maintained by others).  You can make a github repository for them, and even request that they be included on cdnjs for distribution.

Thanks again for your efforts on this.  You can ask here if you have questions about how to do a third-party extension like this.

Quuxplusone:
On top of all those reasons, I subsequently figured out that my use-case (a Jekyll blog using Kramdown) already handled "math blocks" in a different way that meant I didn't need to use any preprocessor at all!  [(See here.)](https://quuxplusone.github.io/blog/2018/08/05/mathjax-in-jekyll/) So I'm okay with abandoning this patch.

dpvc:
OK, thanks for the update.  Sorry you had to do the extra work, but glad you have something that does the job.  And thanks again for offering your code originally!

Issue #1770, opened 2017-06-15
Bug: underscore not handled correctly in \text{} macro


pkienzle:
### Issue Summary

With the amsmath package, LaTeX raises an error on `\text{underscore_separated}` unless there is a leading backslash before the underscore.  Mathjax renders with and without the backslash, but shows the backslash if it is included.  This means that you cannot use the same input for both html and latex markup.

### Steps to Reproduce

1. paste the following into the mathjax live demo page

    erfc is the complementary error function $1 - \text{error\_function}(x)$.

2. Compare this to the same markup in LaTeX (via overleaf shared latex editor):

    https://www.overleaf.com/read/xdcwcrkkwkns

### Technical details

* MathJax Version: https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-MML-AM CHTML-full
* Client OS: Windows, Mac
* Browser: Firefox, Chrome, IE, Safari
*

pkra:
That is expected behavior, cf. http://docs.mathjax.org/en/latest/tex.html#mathjax-tex-and-latex-support.

dpvc:
See also [this forum post](https://groups.google.com/forum/#!msg/mathjax-users/wSh6-hSIUpQ/KmzZhQQGslgJ) for a possible work-around.

pkienzle:
Seems awkward.  In practice, I should be using \mathrm or \operatorname instead so that the fonts are correct, and these can be done with the same syntax in latex (both with the backslash before the underscore).

On further investigation of font issues, the following renders using the same font in latex, but on the mathjax demo page (OS/X firefox) the first character is serif and the second is sans serif:

    $\text{i}$i

The first character renders with style="font-family: STIXGeneral;" in the text span, and style="... font-size: 120%;" in the $$ span. The second character inherits font-family and font-size directly from body.  Latex renders these both with \familydefault.

dpvc:
[...]

Issue #1758, opened 2017-05-31
\textit nested in \texttt (or vice versa) not handled


Eelis:
Consider:

![screenshot](http://eel.is/mathjax-texttt.png)

pkra:
That's expected behavior. Only $ is recognized within `\text{}`, cf. http://docs.mathjax.org/en/latest/tex.html?highlight=exception

Eelis:
Hm, does that mean it's considered too hard to fix? Or is that oversimplifying the case?

Eelis:
Because I thought a typical use case for MathJax was to take formulas from LaTeX documents and feed them through MathJax to then publish them on the web. That's how I've been using it, and the document I happen to be most interested in (the C++ ISO standard) uses this textit/texttt nesting inside math formulas. :)

Eelis:
So, closed without any explanation of why this functionality is excluded and why it is allowed to undermine typical use cases? Wow, what a toxic project.

pkra:
It would help to follow our suggestions at

https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md#reporting-an-issue 

which is linked to from the top of any new issue at submission. In particular, please note https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md#conduct.

> So, closed without any explanation of why this functionality is excluded and why it is allowed to undermine typical use cases?

This is not a typical use case for MathJax, I'm afraid. 

There are many discussion here on the tracker as well as [on the user group](https://groups.google.com/forum/#!searchin/mathjax-users/text$20mode$20macro%7Csort:relevance) around text mode macros. 

But for the benefit for future viewers of this issue: There are two main reasons for text-mode macros are not supported. 

a) TeX's text mode is an entirely different beast from math mode and cannot be implemented easily. 

b) MathJax's internal format is (essentially) MathML which is limited when it comes to text mode content. In particular, it cannot adequately model the output one would expect from something as simple as `\texttt{some \textit{important} text }`.

Issue #1736, opened 2017-04-17
one-character macros inside \text{}


Alex-Jordan:
I discovered the following with the percent sign in an MBX project. I investigated a little and the issue extends to other characters.

With:
```
$\text{\&}$
$\text{\%}$
$\text{\#}$
$\text{\_}$
$\text{\{}$
$\text{\}}$
$\text{\$}$
```

the first four render with the backslash present (as in they appear like `\%` instead of just `%`). The last three do not, which is the right behavior. I checked that a LaTeX document will render all of these without the backslash. I did my MathJax testing at math.se.

pkra:
This is expected behavior as MathJax does not support text mode macros (except $ to switch back to math mode). See http://docs.mathjax.org/en/latest/tex.html?highlight=text-mode

Alex-Jordan:
Ah, thanks.

Issue #1473, opened 2016-05-14
LaTeX input: \% is interpreted as literal \%, not as a command printing a %. 


gunterkoenigsmann:
Dear all,

One half-way minimal example:
```
<!DOCTYPE HTML>
<HTML>
  <HEAD>
    <script type="text/x-mathjax-config">
      MathJax.Hub.Config({
      displayAlign: "left",
      context: "MathJax"
      })
    </script>
    <script type="text/javascript"
	    src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML">
    </script>
    <link rel="stylesheet" type="text/css" href="TableForm_htmlimg/TableForm.css"/>
  </HEAD>
  <BODY>
    \[
    \tag{\%{}t1}\label{t1} \begin{align}1 & 2\\
    3 & 4\end{align}
    \]
  </BODY>
</HTML>
```
The label that is shown in the browser should be %t1, not \%t1.

dpvc:
The contents of the `\tag{}` command is text-mode material.  MathJax does not process macros in text-mode, only math-mode, so you should use `\tag{%t1}` in this case.  (Your `\tag{%{}t1}` produces `\%{}t1`, not just `\%t1`.)  See #790, #569, #346, #243, and others.

PR #2051, opened 2018-09-03
Add extension to process some text-mode macros

dpvc:
This PR creates an extension that allows `\text{}` and other similar macros to process text-mode macros within their argument.  This has been a long-standing request; see #1736, #1758, #1770, #1473, #243, #346.

### DIFF
--- a/unpacked/extensions/TeX/text-macros.js
+++ b/unpacked/extensions/TeX/text-macros.js
@@ -1,0 +1,487 @@
+/*************************************************************
+ *
+ *  MathJax/extensions/TeX/text-macros.js
+ *  
+ *  Implements the processing of some text-mode macros inside
+ *  \text{} and other text boxes.
+ *
+ *  ---------------------------------------------------------------------
+ *  
+ *  Copyright (c) 2018 The MathJax Consortium
+ * 
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


pkra:
Would it be ok to move this into an independent extension?

pkra:
> Would it be ok to move this into an independent extension?

To clarify, I don't mean not to have this in the core. I'm suggesting something like the process for the mhchem extension.

dpvc:
> > Would it be ok to move this into an independent extension?
>
> To clarify, I don't mean not to have this in the core. I'm suggesting something like the process for the mhchem extension.

Are you saying you would like to maintain this extension?  I'm not sure quite what you are suggesting.

I've been a little unhappy about the mhchem extension being both part of the core and a separate extension.  This came about because of an historical accident:  I had implemented a (simplified) work-alike for the mhchem extension, which was put into the core, but later the original author asked to rewrite it, and that was done as a separate extension.  But because my original version was in core, we needed to maintain it there as well.  This has lead to some difficulties in keeping things up to date.

If you are looking to make this a third-party extension, I could see that as a possibility, but I'm leery of another mixed third-party-in-core arrangement.

pkra:
The advantage of maintaining extensions in separate repositories is that they can have releases independently of the (very slow) core. For example, there are technically 5 open PRs about extensions waiting for attention at this time.

Also, I'm also not suggesting a "third party" repository, just a separate one so that this can at least be pulled in via npm with github references.

dpvc:
OK, I see what you are saying.  I will talk to Volker about it and see what we come up with.

Of course, because we are working hard on v3, the v2 updates have been unusually delayed.  I expect that once v3 is out, updates will be more timely.

Issue #2055, opened 2018-09-19
Include complete working example with server script


cottrell:
i.e.

    cd test
    live-server

or whatever. Can have any dependency. 

Would be good to hit failure faster with a script. I am finding it fails without error.

cottrell:
See #2055

PR #2056, opened 2018-09-19
Add test example to show how to test locally. Related to issue #2055

cottrell:


dpvc:
I'm not sure I understand what issue you are trying to address, here.  Can you explain what the purpose is?

Also, you seem to indicate that "the wiki and note should point to this", but can you indicate which links you are talking about?

cottrell:
A quick simple test is to run live-server (or your choice) in the test folder. This produces errors for me and shows MathJax not working when I visit the local site. The example in simple_test works since the inclusion is simpler (it is using a cdn however).

I am a big fan of singleton's for quickstart tests and usage examples. Meant to help the next newcomer that is all.

PR #2109, opened 2019-01-07
Add packed version of mhchem 3.3.0

drify:
In https://github.com/mathjax/MathJax/commit/65685f43415ec96ef43d9038e40ec25abcb5fc22 mhchem extension under the unpacked/ directory has been updated, but the packed version hasn't yet. As a consequence, if `legacy: false` is specified, we don't get the latest version.

This version was copied from [MathJax-mhchem/mhchem.js](https://github.com/mhchem/MathJax-mhchem/blob/master/mhchem.js)


drify:
I'm sorry I had chosen the wrong branch. I just read [CONTRIBUTING.md](https://github.com/mathjax/MathJax/blob/master/CONTRIBUTING.md) and use the correct base branch now. I hope this problem would be noticed soon.

PR #2208, opened 2019-10-02
Fix some README typos & spelling

stakx:
I would've additionally normalized the spelling of `node`, `NodeJS`, etc. to `Node.js`, however the former spellings are used in several MathJax repos so I thought it best to leave those as is for now.


dpvc:
Thanks for the fixes.

Issue #2242, opened 2019-11-18
Use a strict version on the mathjax-full dependency


Mogztter:
Currently the `package.json` file is using a git URL:

https://github.com/mathjax/MathJax/blob/fb0d28b5fa6fe6112c25d11ba8b5f6964e700559/package.json#L53-L55

Using git protocol can be an issue for people behind a corporate proxy and I think it's better to use a fixed version rather than using the master branch.
Also do we really need the sources as a dependency? What is the reasoning behind it? Maybe as `devDependencies` but if I just want to use MathJax, this module should be enough right?

dpvc:
You are right, this should just be a dev dependency.  It is only needed for building the distribution files that are part of the final mathjax npm package.  I'll update the `package.json` file.  Thanks for the catch!  There is certainly no need for `mathjax-full` to be installed for people using the `mathjax`.

dpvc:
I've made a pull request for this, and it will be included in the next release.

Mogztter:
Excellent news, thanks @dpvc 👍

PR #2244, opened 2019-11-19
Move mathjax-full to a dev-dependency.  #2242

dpvc:
Change `mathjax-ful`l dependency to a devDependency, so `mathjax-full` (and its dependencies) are not loaded when `mathjax` is included in a node project.

Resolves issue #2242

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "make-es5": "npm run --silent install:all && npm run --silent compile && npm run --silent components && npm run --silent move",
     "postmake-es5": "npm run --silent title --mathjax:title='Cleaning Up...' && npm run --silent clean:node"
   },
-  "dependencies": {
+  "devDependencies": {
     "mathjax-full": "git://github.com/mathjax/MathJax-src.git"
   }
 }
--- a/package.json
+++ b/package.json
@@ -12,6 +12,9 @@
     "browser",
     "node"
   ],
+  "devDependencies": {
+    "mathjax-full": "3.0.0"
+  },
   "maintainers": [
     "MathJax Consortium <info@mathjax.org> (http://www.mathjax.org)"
   ],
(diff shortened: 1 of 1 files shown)


Mogztter:
LGTM :wink:

chiptus:
can mathjax-full be loaded from npm instead of git?

Mogztter:
> can mathjax-full be loaded from npm instead of git?

That would be even better since the package is published on npmjs: https://www.npmjs.com/package/mathjax-full

But it might change the release process a little bit because you would have to:

1. release `mathjax-full` (for instance version 3.0.1)
2. edit the package.json in this repository to update the version of `mathjax-full` (to version 3.0.1)
3. release `mathjax`

chiptus:
you can pin the version i.e `mathjax-full: "3"`, no? sounds even better than using the git repo as it's not pinned

chiptus:
quick fix for people who need this to be loaded from the npm repo and are using yarn:
add this to your package.json:
```
{
  ...other package.json things,
  "resolutions": {
     "mathjax-full": "3"
  }
}
```

Mogztter:
@chiptus Do you know if there's a workaround for npm?

chiptus:
Maybe https://github.com/rogeriochaves/npm-force-resolutions

dpvc:
I've updated to use the mathjax-full package, as @Mogztter suggested.  The `es5` directory is already included in that package, so we can just copy that one rather than building it directly.  The `package.json` has to be edited to update its version anyway, so it is no problem to adjust the mathjax-full version as well at that time.

We are preparing a v3.0.1 release, and these changes will be used for that version.

Mogztter:
Excellent ! Thanks @dpvc

chiptus:
Thanks!

PR #2645, opened 2021-03-09
[es6] Target a newer standard of javascript

artagnon:
See mathjax/MathJax-src#616 for more.


dpvc:
This repository is set up to populate the CDNs that serve MathJax.  The changes to this repository are scripted, and come from the `mathjax-full` npm package (which in turn comes from the MathJax-src repository).  We don't accept direct PRs to this repository for version 3 (and v2 is not being updated except for maintenance).  When we start to include an ES6 version, the scripts for updating this repository will need to be modified so that the appropriate files are included automatically, and so that they become part of the `mathjax` npm package.

For that reason, I'm closing this PR.

PR #2763, opened 2021-09-02
CONTRIBUTING: Fix broken CSS-Tricks link

EricDunsworth:
The "reduced test case" link in the bug reports section was pointing to a 404 error page.

This brings it back to life.

Port of mathjax/MathJax-src#740


dpvc:
Thanks again.

PR #2809, opened 2021-12-03
[config] Fix menu label of Traditional Chinese

toto6038:


toto6038:
Current label for Traditional Chinese on language selection menu is "汉语", which is neither in Traditional Chinese characters nor a correct name. The PR fixes that issue and also improve consistency on the language names.

toto6038:
@dpvc This PR deals the same problem with [that](https://github.com/mathjax/MathJax-src/pull/751/files), and follows [your suggestion](https://github.com/mathjax/MathJax-src/pull/751#issuecomment-984770142) to move here.

shun2wang (approved):
Yes, that's  makes sense.

